Issue #948, opened 2021-07-03
documentation: Modify the confusing tour/references/aliases page


cueckoo:
_Originally opened by @PierreR in https://github.com/cuelang/cue/issues/948_

**Documentation of aliases & let**

The following page of the doc refering to `aliases` is using the `let` construct:
https://cuelang.org/docs/tutorials/tour/references/aliases/

But to be honest I am not sure if the page is meant to introduce the concept of `aliases` or the `let` construct. These two are treated in the spec.md as two different concepts (with different use cases).

**Describe the solution you'd like**

I suggest to align the page with the spec.md to avoid confusion.

To the best of my knowledge a more idiomatic way to write the code snippet of the aliases page would be:
```
A=a: {
    d: 3
}
b: {
    a: {
        // A provides access to the outer "a" which would
        // otherwise be hidden by the inner one.
        c: A.d
    }
}
```

cueckoo:
_Original reply by @mpvl in https://github.com/cuelang/cue/issues/948#issuecomment-837922840_

This is an artifact of the time when there was no `let` and the same syntax/terminology was used to describe both. 

This is indeed very misleading and should be fixed.

mvdan:
The latest version of https://cuelang.org/docs/tour/references/aliases/ does not make use of any `let` declarations, so I believe this issue is fuxed.

PR #1098, opened 2021-07-14
doc/tutorials/basics: fix example of aliases

yujunz:


myitcv:
Thanks for re-raising this PR here, and apologies for not being able to transfer it automatically.

As I mentioned in the original issue, and per @mpvl's comment in the original PR, I think we need to first conclude in https://github.com/cue-lang/cue/issues/948 exactly what changes we want to make. Because, as the issue suggests, we likely want different pages for aliases and let. Then we can revisit the changes you have proposed.

mvdan:
Per Paul's comment above, and also due to the conflicts that this has accumulated over time, I'm going to close this PR for now. The issue remains open to improve the tutorial. Thank you again for contributing, @yujunz!

PR #1105, opened 2021-07-15
ci: improve goreleaser workflow

### LINKED ISSUE #1100: releases: switch to using proxy mode of goreleaser
In https://github.com/goreleaser/goreleaser/pull/2129, @caarlos0 added support to `goreleaser` for building release assets from proxy-resolved modules, the benefit being that resulting binaries include full module fidelity information and sums from sum.golang.org. 

We should update the CUE `goreleaser` configuration to use that new setting.

caarlos0:
closes #1100

- updated to latest goreleaser 
- using gomod proxy
- improve version command to use build info only:
  - on on proxy builds (e.g. `go run`) it will show `cue version devel` just as before
  - on proxied builds it will show the mod version + its checksum


.github/workflows/release.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       uses: goreleaser/goreleaser-action@v2
       with:
         args: release --rm-dist
-        version: v0.155.1
+        version: v0.173.2
```
myitcv, line 38: This change actually needs to be made in the source CUE, which is why CI is failing.

It's easily missed, but this file is generated, per the header:

https://github.com/cue-lang/cue/blob/7d25ee232fe7ee57c502baee7e82417a03a7e297/.github/workflows/release.yml#L1

So the change ends up needing to be made here:

https://github.com/cue-lang/cue/blob/7d25ee232fe7ee57c502baee7e82417a03a7e297/internal/ci/workflows.cue#L243

After making that change, you need to regenerate the workflow files:

```
go generate ./internal/ci
```

As if this process weren't involved enough already, we actually have tests that are based on our workflow setup (which drastically reduces the iteration time when working on changes to the core evaluation engine). Those tests include golden files that are a the result of generating Yaml from CUE (amongst a number of others). To update those golden files, run:

```
CUE_UPDATE=1 go test ./...
CUE_UPDATE=1 go test ./...
```

Note that I have deliberately run that command twice, above.

Then a plain:

```
go test ./...
```

should succeed, and:

```
go generate ./internal/ci
```

should at that point leave you with a clean commit.

.goreleaser.yml
```diff
@@ -4,15 +4,18 @@ before:
   hooks:
     - go mod download
 
+gomod:
+  proxy: true
```
myitcv, line 4: Very excited for this :)

myitcv (commented):
@caarlos0 thanks so much for helping out with this. I definitely feel much more comfortable knowing that you have looked over our setup 😄

.github/workflows/release.yml
caarlos0: oh, totally missed that part... fixing!

caarlos0:
also just noticed you are not using goreleaser to create the docker images, any particular reason? Looking at it it seems it would be doable through goreleaser as well... 🤔

caarlos0:
OK, fixed what you pointed out, and also:

- fixed an invalid tag `version` in go setup
- using goreleaser to build docker images
- creating a docker manifest with both arm64 and amd64 images

myitcv:
@caarlos0 this is awesome thanks. 

One final request: please can you squash the commits down to a single commit message, ready for import into Gerrit? I'd also suggest we prefix the first line of the commit message with `ci:`  - correctly categorises it with CI-related changes then (we need to do better at documenting this rather "loose" selection of categories).

.. and one question. How should I be testing this locally? Or better still, how can we automate the testing of this setup? Perhaps not something to trigger on every build matrix entry, but I've been concerned for some time that we don't really trigger this workflow until we come to a release... at which point it's "too late".

caarlos0:
@myitcv squashed and changed message prefix!

---

About your question: you can run `goreleaser release --rm-dist --snapshot`... it will test everything but the publishing part... I can do that as well if you want :)

myitcv:
> you can run `goreleaser release --rm-dist --snapshot`... it will test everything but the publishing part... I can do that as well if you want :)

Ah cool. Let's tackle that in a follow-up PR. That check need only run for the Linux entry of the latest stable Go version, so we can use a guard much like this:

https://github.com/cue-lang/cue/blob/7d25ee232fe7ee57c502baee7e82417a03a7e297/internal/ci/workflows.cue#L87-L89

Unlike that linked step, I think it _does_ make sense to run as part of a non-`master` build. i.e. we do want to check it as part of any contribution. Limiting the check to just Linux and just the latest stable Go version will keep runtimes low.

myitcv (approved):
LGTM - pinging @zchee who kindly did the initial setup for any thoughts?

zchee (approved):
LGTM

myitcv:
Ok - we're all set here... and we've run into the first little issue post migration - I can't import the PR into Gerrit. Strange, because that _definitely_ worked during my testing. So I've raised https://bugs.chromium.org/p/gerrit/issues/detail?id=14807 as a follow up.

@caarlos0 - whilst we wait for a resolution there, I tested things locally using https://github.com/cue-lang/cue/pull/1105#issuecomment-882503089, and found that `go.sum` was modified as a result. The reason being, we currently have `go mod download` as a before hook:

https://github.com/cue-lang/cue/blob/7d25ee232fe7ee57c502baee7e82417a03a7e297/.goreleaser.yml#L5

That seems superfluous to my mind. I removed it and tested again, and all seemed fine.

Is it safe for me to remove this in a follow up?

caarlos0:
> That seems superfluous to my mind. I removed it and tested again, and all seemed fine.
> 
> Is it safe for me to remove this in a follow up?

hmm... usually I set it to `go mod tidy` to make sure there are no changes in build time... but if removing works I think its ok as well...

myitcv:
> hmm... usually I set it to `go mod tidy` to make sure there are no changes in build time... but if removing works I think its ok as well...

That raises an interesting question. Because having `go mod tidy` as a before hook actually means that the `goreleaser` workflow _can_ make changes to `go.{mod,sum}` files. I would have thought it probably best to not have `go mod tidy`, because then any of the build steps that run as part of `goreleaser` would fail if they did need to make changes (this is the new behaviour with Go 1.16, where build commands fail if any of `go.{mod,sum}` is not tidy/complete), thereby alerting the user. As it stands, have `go mod tidy` would mean that these changes are missed, unless the user then checks the commit is clean.

I'll make the change in a follow-up in any case because just so long as I doesn't break `goreleaser`, I think it's the right change for us here.

caarlos0:
hmm, I usually do the `go mod tidy` thing to make sure the `go.*` files are "clean"... if not, goreleaser will change them and fail the build later when it checks for a clean git state...

myitcv:
> hmm, I usually do the `go mod tidy` thing to make sure the `go.*` files are "clean"... if not, goreleaser will change them and fail the build later when it checks for a clean git state...

Interesting, great to hear that `goreleaser` checks that. So is there an issue here then? Because we have `go mod download` as our before step, and that _does_ modify `go.sum`. So why is `goreleaser` not failing in this case?

caarlos0:
anyway, removed the before hook and rebased with master

myitcv:
Ok, I've gone with manually importing this into Gerrit for now. See progress at:

https://review.gerrithub.io/c/cue-lang/cue/+/521069

.github/workflows/release.yml
```diff
@@ -19,45 +19,22 @@ jobs:
...
       with:
-        go-version: ${{ matrix.go-version }}
-        version: 1.15.8
+        go-version: 1.15.8
         stable: false
+    - name: Setup qemu
+      uses: docker/setup-qemu-action@v1
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v1
+    - name: Docker Login
```
myitcv, line 28: I've just spotted an issue whilst testing this locally. The docker images that get created are:

```
      • building docker image     image=docker.io/cue-lang/cue:v0.99.99-next-amd64
      • building docker image     image=docker.io/cue-lang/cue:v0.99.99-next-arm64
```

(I was testing against the fictional `v0.99.99`).

However, our Docker hub repository is `cuelang/cue` - i.e. it did _not_ get migrated.

Is there some way we can configure this?

caarlos0: yup, let me fix it :)

caarlos0: fixed!

PR #1184, opened 2021-08-03
crypto/ed25519: support signature verification

### LINKED ISSUE #331: /pkg/crypto/ed25519 `Valid` support
_Originally opened by @rudolph9 in https://github.com/cuelang/cue/issues/331_

## Summary

Support for validating ed25519 signatures. This would essentially map directly onto the golang [`ed25519.Verify`](https://godoc.org/crypto/ed25519#Verify) method. 

## Considerations
Support for an Ed25519 `Valid` builtin seems to fit well with the hermetic nature of cue and aligns well with the existing crypto packages md5, sha1, etc.

This request is intentionally very limited to NOT requesting support for `GenerateKey` or `Sign`.  1) managing private keys opens a lot of cans of worms and likely outside of the intended scope of cuelang.  2) GenerateKey is not idempotent (unless a seed is provided but again more cans or worms) and I can't remember off the top of my head if given the same `privateKey` and `message` if [`Sign`](https://godoc.org/crypto/ed25519#Sign) is idempotent but regardless; private keys and worms.

## Usage
example usage:
```
import "crypto/ed25519"

publicKey: bytes
message: bytes 
signature: bytes

isValid: ed25519.Verify(publicKey, message, signature)
```

## impl

Probably as simple as
```
package ed25519

import "crypto/ed25519"

func Valid(publicKey, message, signature []byte) bool {
	return ed25519(publicKey.(ed25519.PublicKey), message, signature)
}
```

oncilla:
Add support for ed25519 signature verification of message based on
a public key.

fixes #331

Signed-off-by: Dominik Roos <domi.roos@gmail.com>


oncilla:
I was not sure how to do the testdata file.
If you have a better pattern for generating the golden file, let me know.

verdverm:
re: golden file, there is a way to run the testsuite in update mode (https://github.com/cue-lang/cue/blob/a9b13e823e76986cf8722edf04174f8ee10c525c/internal/cuetxtar/txtar.go#L323) not sure how to invoke that myself though...

Seems like a good topic to add to https://github.com/cue-lang/cue/wiki/Contribution-tips-and-tricks, and generally guidance on writing such tests

oncilla:
Oh, I see. It is done via env var here: https://github.com/cue-lang/cue/blob/a9b13e823e76986cf8722edf04174f8ee10c525c/internal/cuetest/cuetest.go#L56

Hm, but that update does not really fit, because we need to generate valid signatures. But we could get rid of the manual output generation.

Thanks for the hint 👍 
I will play around with.it tomorrow.

oncilla:
I updated the test to use `cuetest.UpdateGoldenFiles`

To update the golden files, you can simply run the following:

```sh
CUE_UPDATE=true go test ./pkg/crypto/ed25519/
```

myitcv:
Thanks @oncilla - @mpvl and I are, independently, taking some vacation at the moment so we will get back to this on our return.

oncilla:
No worries, enjoy!

pkg/crypto/ed25519/ed25519.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+const (
+	// PublicKeySize is the size of a public key in bytes.
+	PublicKeySize = 32
+)
+
+// Verify verifies the provided signature of the message using the public key.
+// An error is returned if and only if an invalid public key is provided.
+func Verify(publicKey, message, signature []byte) (bool, error) {
```
mpvl, line 31: Both the proposal and the convention within CUE is to call this `Valid`, for better or for worse.

mpvl (changes requested):
Sorry for the enormous delay. 

This generally looks good. We have one small tweak that we need to do in Gerrit, also because we can't run CI on GH right now.

myitcv (approved):
LGTM with the caveat @mpvl mentioned in https://github.com/cue-lang/cue/pull/1184#discussion_r810067255

oncilla:
:tada:

Issue #1125, opened 2021-07-21
general: support MacPorts as an installation source


myitcv:
Raising off the back of #1121, to capture discussion about whether we add MacPorts support or not.

I know nothing about MacPorts, so will flesh out this description based on replies below.

**Is your feature request related to a problem? Please describe.**

TBC

**Describe the solution you'd like**

Support for MacPorts as an installation source:

```
sudo port selfupdate
sudo port install cue
```

**Describe alternatives you've considered**

TBC

**Additional context**

n/a

@herbygillot - I would be grateful if you could help to flesh out the points above.

myitcv:
Thanks for raising this @herbygillot. 

We already maintain a homebrew setup at https://github.com/cue-lang/homebrew-tap, managed via [GoReleaser](https://goreleaser.com/). 

I'm not super familiar with MacPorts, so forgive the basic questions:

* How many people use MacPorts?
* Is the Homebrew setup insufficient/inappropriate in some way?
* How do you propose to maintain the MacPorts config file, which I think lives at https://github.com/macports/macports-ports/blob/master/devel/cue/Portfile

herbygillot:
So I'm a volunteer maintainer with the MacPorts project.

In a sense, MacPorts is an alternative package manager for macOS.  It has a different philosophy and mode of operation compared to Homebrew.  Software projects and packages are added to MacPorts as **ports**, which are defined by **Portfiles**.  You are correct that [this](https://github.com/macports/macports-ports/blob/master/devel/cue/Portfile) is the Portfile for `cue`.

> How many people use MacPorts?

I can't say for sure - there are statistics, but those are from advanced users who knowingly opt-in to have their stats uploaded.

> Is the Homebrew setup insufficient/inappropriate in some way?

As mentioned before, MacPorts is an alternative to Homebrew.  Some folks use Homebrew to manage open source software on their Mac, others use MacPorts, and still some other adventurous folk use both.  Adding `cue` to the MacPorts project in no way implies some insufficiency with Homebrew, it's just making `cue` available to people who prefer or primarily use MacPorts.

> How do you propose to maintain the MacPorts config file

Portfiles are kept up-to-date by maintainers.  In most cases, Postfiles are marked `openmaintainer`, meaning that anyone can open a pull request to update a Portfile, which is then approved by the maintainer or MacPorts administrators.  This is indicated within the `maintainers` section in the Portfile.  But the primary responsibility of keeping a port current lies with the maintainer.

[...]

PR #1241, opened 2021-09-01
docs: migrate to homebrew-core

chenrui333:
relates to https://github.com/Homebrew/homebrew-core/pull/84359

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ Some aspects of this are:
 
 Using [Homebrew](https://brew.sh), you can install using the CUE Homebrew tap:
 
-`brew install cuelang/tap/cue`
+`brew install cue`
 
 #### Install from Source
 
@@ -179,4 +179,3 @@ Unless otherwise noted, the CUE source files are distributed
 under the Apache 2.0 license found in the LICENSE file.
 
 This is not an officially supported Google product.
-


verdverm:
related: https://github.com/cue-lang/cue/issues/1125

The same question would be how does this ensure the same binary is produced? I suspect that it cannot be guaranteed due to inability to ensure the same Go version is used.

See how the cue tap downloads the binaries: https://github.com/cue-lang/homebrew-tap/blob/master/cue.rb

There will also be the issue of effort to maintain in CUE's repos vs making PR's against Homebrew's. @myitcv has setup automation for the release process. Note that the tap is actually updated by the Goreleaser tool. https://github.com/cue-lang/cue/blob/master/.goreleaser.yml#L54

It is generally prefered to open an issue or discussion before opening a PR. Having `cue` unofficially in the homebrew tap while the official version is maintained through the cue tap may be an undesirable situation.

chenrui333:
homebrew-core formulae is pretty well maintained. And the contribution to homebrew-core is very easy.

There are also couple of options of automate the version bump process as well, like:
- https://github.com/mislav/bump-homebrew-formula-action
- https://github.com/dawidd6/action-homebrew-bump-formula

And homebrew-core is also working on automate this process as well.

verdverm:
@chenrui333 It looks like homebrew does not like core formulas downloading binaries. Is that an accurate assessment of the preferences?

Per #1125 concerns, how can the formula ensure that the binaries have the same hash as the official releases when compiling ? My understanding is that the same Go version 1.x.y would need to be used.

chenrui333:
> how can the formula ensure that the binaries have the same hash as the official releases when compiling ?

I think if you can always ensure the build use the latest go to build and use it for both homebrew-core ass well as in here, this should be out of concern.

Also we are not shipping software without responsibility, meaning, we can always patch the build and update if needed. (that being said, if there is any usage issue, we can troubleshot and patch the artifact)

Does that make sense?

verdverm:
Based on this file: https://github.com/Homebrew/homebrew-core/pull/84359/files

The Go version does not appear to be locked, so `cue` would be compiled with whichever version the user has on their system at the time of installing `cue` via homebrew.

`cue-v0.4.0` is compiled with Go 1.15.8 (https://github.com/cue-lang/cue/blob/master/.github/workflows/release.yml#L22)

So if a user has the latest Go (v1.17.0 at the time of writing) then they will have a different binary. This seems like it will happen more often than not, considering Go releases patches pretty frequently. i.e. The latest Go version will be different from the one used to compile more days than it matches, and subsequently the homebrew installed version would be different from the official binary more days than not. This is why the official `cue` tap installs by downloading the binary rather than compiling.

Does this help clarify the same concern raised in #1125 ?

I actually don't think that the "using goreleaser to build would result in the same binary" holds. While it will guarantee the same dep hashes are embedded into the binary for provenance, the Go binary is paired with the standard library, which changes between versions. Thus it seems there will be different code being compiled.

I suppose the formula could fetch the matching Go version, compile with that, and then remove it. Seems more complicated than fetching the official binaries. Is it possible for a homebrew-core formula to fetch a binary rather than compiling from source?

Issue #1109, opened 2021-07-17
spec: detail which parts of the spec are yet to be implemented


myitcv:
Whilst we await #1013, the spec ends up being one of the main places people reference for language features. 

In https://github.com/cue-lang/cue/issues/1065 and on Slack, someone asked why `...string` did not work. The reason is that it is not yet implemented.

In the interim it seems prudent to annotate such unimplemented features so the reader's expectation is set correctly. This also acts as something of a TODO list as maintainers.

mvdan:
Worth noting that now we have https://github.com/cue-lang/cue/issues?q=is%3Aissue%20state%3Aopen%20label%3Aspec-deviation to track issues where the spec and implementation do not agree; my impression is that we can close this issue as long as all such instances are either tracked in the issue tracker with the label, or marked as "notes" in the spec text directly, just like `...string`.

myitcv:
> my impression is that we can close this issue as long as all such instances are either tracked in the issue tracker with the label, or marked as "notes" in the spec text directly, just like `...string`.

I think the spec, which is read by a non-zero number of people, needs to be clear where things are not yet implemented. Such items should also be tracked here. i.e. we need to call things out in the spec (for the reader) and in the issue tracker (for people to follow when a feature is implemented). Only tracking something here doesn't help the reader of the spec.

PR #1268, opened 2021-09-22
spec: Update example with regards to pattern constraints

slewiskelly:
The pattern constraint example is slightly incorrect and does not constrain
fields as it is supposed to.

The default constraint example also does not work, but this is not yet
implemented, see https://github.com/cue-lang/cue/issues/1109.

Signed-off-by: Scott Lewis-Kelly <slewiskelly@slewiskel.ly>

### DIFF
--- a/doc/ref/spec.md
+++ b/doc/ref/spec.md
@@ -1091,10 +1091,10 @@ The token `...` is a shorthand for `..._`.
 
 ```
 a: {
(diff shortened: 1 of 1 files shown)


slewiskelly:
Closing in favour of https://review.gerrithub.io/c/cue-lang/cue/+/525798.

PR #1288, opened 2021-10-07
Fix doc string for CompileBytes and references

emcfarlane:
Simple doc fix for `CompileBytes` function and docs that refers to it.


myitcv (approved):
LGTM and sincere apologies for letting this lie untouched for so long.

PR #1301, opened 2021-10-15
Add CIDR range verification function in net package

jared-gs:
I've added a CIDR verification function, which helps in use-cases where CIDR ranges (and not individual IP addresses) need to be checked.

### DIFF
--- a/pkg/net/ip.go
+++ b/pkg/net/ip.go
@@ -75,6 +75,36 @@ func netGetIP(ip cue.Value) (goip net.IP) {
 	}
 }
 
+func netGetIPCIDR(ip cue.Value) (gonet *net.IPNet) {
+	switch ip.Kind() {
+	case cue.StringKind:
+		s, err := ip.String()
+		if err != nil {
+			return nil
+		}
+		_, gonet, err := net.ParseCIDR(s)
+		if err != nil {
+			return nil
+		}
+		return gonet
+
+	case cue.BytesKind:
+		b, err := ip.Bytes()
+		if err != nil {
+			return nil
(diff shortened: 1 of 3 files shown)


pkg/net/ip.go
```diff
@@ -112,6 +142,14 @@ func IP(ip cue.Value) bool {
 	return netGetIP(ip) != nil
 }
 
+// IPCIDR reports whether s is a valid IPv4 or IPv6 address with CIDR subnet notation
```
seh:
```suggestion
// IPCIDR reports whether ip is a valid IPv4 or IPv6 address with CIDR subnet notation.
```

jared-gs: Updated, thanks. I will note other functions in this package will need similar edits.

pkg/net/ip.go
```diff
@@ -112,6 +142,14 @@ func IP(ip cue.Value) bool {
 	return netGetIP(ip) != nil
 }
 
+// IPCIDR reports whether ip is a valid IPv4 or IPv6 address with CIDR subnet notation.
+//
+// The address may be a string or list of bytes.
+func IPCIDR(ip cue.Value) bool {
+	// TODO: convert to native CUE.
+	return netGetIPCIDR(ip) != nil
```
mpvl, line 150: CUE can pass error messages back to the user if you use the `IPCIDR(ip cue.Value) (bool, error)` signature. I know it looks strange, but I would do that here for the sake of better error messages.

mpvl (changes requested):

PR #1385, opened 2021-11-18
cue: fix output in ExampleContext

### LINKED ISSUE #1267: The testable example "ExampleContext" is not testing output
👋  Hello, 

I am investigating the golang issue: https://github.com/golang/go/issues/48362, where testable examples are causing false positives when the output comment block is not the last comment block example.

I ran a parser against the [top golang repos](https://github.com/kaxap/arl/blob/master/README-Go.md), including this repo, and I found this bug with your [ExampleContext testable example](https://github.com/cue-lang/cue/blob/e8550b89779d293bbc9b530d2c4c9f9cf4c6f3d0/cue/examplecompile_test.go#L24-L55).

The output comment blocks at the end should be combined like so:

```
func ExampleContext() {
	ctx := cuecontext.New()

	v := ctx.CompileString(`
		a: 2
		b: 3
		"a+b": a + b
	`)

	p("lookups")
	p("a:     %v", v.LookupPath(cue.ParsePath("a")))
	p("b:     %v", v.LookupPath(cue.ParsePath("b")))
	p(`"a+b": %v`, v.LookupPath(cue.ParsePath(`"a+b"`)))
	p("")
	p("expressions")
	p("a + b: %v", ctx.CompileString("a + b", cue.Scope(v)))
	p("a * b: %v", ctx.CompileString("a * b", cue.Scope(v)))

	// Output:
	// lookups
	// a:     2
	// b:     3
	// "a+b": 5
	//
	// expressions
	// a + b: 5
	// a * b: 6
}
```

I am happy to PR this in, but it will take a few days for me to get approval from my employer before I can submit it.

ameowlia:
If the output comment block of a testable example is not the last comment block then nothing will be tested. This commit combines the comment blocks so that the example actually tests against the output.

fixes #1267

### DIFF
--- a/cue/examplecompile_test.go
+++ b/cue/examplecompile_test.go
@@ -44,7 +44,7 @@ func ExampleContext() {
 	// a:     2
 	// b:     3
 	// "a+b": 5
-
+	//
 	// expressions
 	// a + b: 5
 	// a * b: 6


myitcv (approved):

myitcv:
Thanks for the fix, @ameowlia!

PR #1390, opened 2021-11-19
WIP: tools/flow: return controller value

eonpatapon:
This can be used to validate the value after the flow is completed, or
unifying the result value with the original value.

Change-Id: If588467ee85d061cce51fcbef2ab8c0fcdd3ead8
Signed-off-by: Jean-Philippe Braun <eon@patapon.info>

### DIFF
--- a/cmd/cue/cmd/custom.go
+++ b/cmd/cue/cmd/custom.go
@@ -131,7 +131,7 @@ func doTasks(cmd *Command, typ, command string, root *cue.Instance) error {
 
 	c := flow.New(cfg, root, newTaskFunc(cmd))
 
-	err := c.Run(context.Background())
+	_, err := c.Run(context.Background())
 	exitIfErr(cmd, root, err, true)
 
 	return err
--- a/tools/flow/flow_test.go
+++ b/tools/flow/flow_test.go
@@ -81,7 +81,7 @@ func TestFlow(t *testing.T) {
 		c := flow.New(cfg, v, taskFunc)
 
 		w := t.Writer("errors")
-		if err := c.Run(context.Background()); err != nil {
+		if _, err := c.Run(context.Background()); err != nil {
 			cwd, _ := os.Getwd()
 			fmt.Fprint(w, "error: ")
 			errors.Print(w, err, &errors.Config{
@@ -209,7 +209,7 @@ func TestX(t *testing.T) {
 
 	t.Error(mermaidGraph(c))
 
-	if err := c.Run(context.Background()); err != nil {
+	if _, err := c.Run(context.Background()); err != nil {
 		t.Fatal(errors.Details(err, nil))
 	}
 }
--- a/tools/flow/flow.go
+++ b/tools/flow/flow.go
@@ -219,12 +219,12 @@ func New(cfg *Config, inst cue.InstanceOrValue, f TaskFunc) *Controller {
 }
 
 // Run runs the tasks of a workflow until completion.
-func (c *Controller) Run(ctx context.Context) error {
+func (c *Controller) Run(ctx context.Context) (cue.Value, error) {
 	c.context, c.cancelFunc = context.WithCancel(ctx)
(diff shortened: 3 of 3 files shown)


eonpatapon:
This is a proposal to change the API of `flow.Run` to return the value updated by the flow.

I have a use case where I want to run multiple flows, and I want to validate resulting flow value.
Moreover some flow might depend on values produced by a previous flow so I need to use values produced by the previous flow.

tools/flow/example_basic_test.go
```diff
@@ -4,12 +4,15 @@ import (
...
 	"log"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
 
 	"cuelang.org/go/cue"
 	"cuelang.org/go/tools/flow"
 )
 
-func Example() {
```
verdverm: Renaming this function from `Example*` to `Test*` removes the example from Godocs. You should probably create a new `Test*` function.

cmd/cue/cmd/custom.go
```diff
@@ -131,7 +131,7 @@ func doTasks(cmd *Command, typ, command string, root *cue.Instance) error {
 
 	c := flow.New(cfg, root, newTaskFunc(cmd))
 
-	err := c.Run(context.Background())
+	_, err := c.Run(context.Background())
```
verdverm, line 134: should the returned value be checked for an error here? (`rv.Err() != nil`)

tools/flow/flow_test.go
```diff
@@ -81,7 +81,7 @@ func TestFlow(t *testing.T) {
 		c := flow.New(cfg, v, taskFunc)
 
 		w := t.Writer("errors")
-		if err := c.Run(context.Background()); err != nil {
+		if _, err := c.Run(context.Background()); err != nil {
```
verdverm, line 84: same should `rv.Err()` checked

tools/flow/flow_test.go
```diff
@@ -209,7 +209,7 @@ func TestX(t *testing.T) {
 
 	t.Error(mermaidGraph(c))
 
-	if err := c.Run(context.Background()); err != nil {
+	if _, err := c.Run(context.Background()); err != nil {
```
verdverm, line 212: same should `rv.Err()` be checked

tools/flow/example_basic_test.go
```diff
@@ -25,12 +28,14 @@ func Example() {
...
 	controller := flow.New(nil, inst, ioTaskFunc)
-	if err := controller.Run(context.Background()); err != nil {
+	rv, err := controller.Run(context.Background())
+	if err != nil {
 		log.Fatal(err)
 	}
-	// Output:
-	// setting a.output to "hello world"
-	// setting b.output to "hello hello world"
+	aout, _ := rv.LookupPath(cue.ParsePath("a.output")).String()
```
verdverm: should these errors be checked rather than ignored, for completeness?

tools/flow/example_basic_test.go
```diff
@@ -25,12 +28,14 @@ func Example() {
 		log.Fatal(err)
 	}
 	controller := flow.New(nil, inst, ioTaskFunc)
-	if err := controller.Run(context.Background()); err != nil {
+	rv, err := controller.Run(context.Background())
```
verdverm: `rv.Err()` should probably be checked as well

verdverm (commented):
@eonpatapon Generally speaking, a proposal should start as a discussion or issue, not with a PR.

w.r.t. this change, can the final value not be obtained from the existing implementation? In the following example, once the last task has call `t.Fill`, the TaskFactory prints the "final" value. One could capture this value by copying it to a variable scoped outside of the function.

https://cuetorials.com/go-api/workflows/custom/

tools/flow/example_basic_test.go
eonpatapon: Yes I realized this later on, I will change this

PR #1419, opened 2021-12-08
Allow hidden tasks in cue/flow

jlongtine:
Allow hidden fields contain tasks would be very useful for us on the dagger team. 

This PR adds a configuration field for `cue/flow` to enable hidden fields as tasks, and includes a small test.

### DIFF
--- a/tools/flow/example_basic_test.go
+++ b/tools/flow/example_basic_test.go
@@ -33,6 +33,41 @@ func Example() {
 	// setting b.output to "hello hello world"
 }
 
+func ExampleHidden() {
+	var r cue.Runtime
+	inst, err := r.Compile("example.cue", `
+	a: {
+		input: "world"
+		output: string
+	}
+	b: {
+		input: a.output
+		output: string
+	}
+	_c: {
+		input: b.output
+		output: string
+	}
+	d: {
+		input: _c.output
+		output: string
+	}
+	`)
+	if err != nil {
+		log.Fatal(err)
+	}
+	controller := flow.New(&flow.Config{FindHiddenTasks: true}, inst, ioTaskFunc)
+	// controller := flow.New(nil, inst, ioTaskFunc)
+	if err := controller.Run(context.Background()); err != nil {
+		log.Fatal(err)
+	}
+	// Output:
+	// setting a.output to "hello world"
+	// setting b.output to "hello hello world"
+	// setting _c.output to "hello hello hello world"
+	// setting d.output to "hello hello hello hello world"
+}
+
 func ioTaskFunc(v cue.Value) (flow.Runner, error) {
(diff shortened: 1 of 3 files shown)


verdverm:
Would it be better to allow setting or passing an options slice for full configurability? What if another user wants to find tasks in definitions or optional values?

It seems best to allow the options to be passed, which would be a stable API, should the options change or expand.

jlongtine:
@verdverm I'm not sure how that would play in this case. 

These are the Options available:
- [`Final()`](https://github.com/cue-lang/cue/blob/master/cue/types.go#L2025)
- [`Schema()`](https://github.com/cue-lang/cue/blob/master/cue/types.go#L2035)
- [`Concrete(concrete bool)`](https://github.com/cue-lang/cue/blob/master/cue/types.go#L2045)
- [`DisallowCycles(disallow bool)`](https://github.com/cue-lang/cue/blob/master/cue/types.go#L2060)
- [`Raw()`](https://github.com/cue-lang/cue/blob/master/cue/types.go#L2084)
- [`All()`](https://github.com/cue-lang/cue/blob/master/cue/types.go#L2090)
- [`Docs(include bool)`](https://github.com/cue-lang/cue/blob/master/cue/types.go#L2100)
- [`Definitions(include bool)`](https://github.com/cue-lang/cue/blob/master/cue/types.go#L2108)
- [`Hidden(include bool)`](https://github.com/cue-lang/cue/blob/master/cue/types.go#L2116)
- [`Optional(include bool)`](https://github.com/cue-lang/cue/blob/master/cue/types.go#L2125)
- [`Attributes(include bool)`](https://github.com/cue-lang/cue/blob/master/cue/types.go#L2130)

The only Options that make sense to me in this context are `Hidden` and `Definitions`. Perhaps `All`?

Having a `Optional`, `Attributes`, or `Docs` task doesn't compute with my mental model of tasks. Am I missing something here? 

@myitcv @mpvl Any thoughts?

verdverm:
@jlongtine My thought is to have an `Options []cue.Option{}` on the config object, rather than making something specific to Hidden. This is more flexible (developer configurable), concise, and future proofs any new options that may be made available. This is the same idea I think should be applied to the `Walk` function (i.e. https://cuetorials.com/go-api/basics/traversing/#custom-walk) and generally areas where Fields is called internally.

RE: recent slack message, a few thoughts
- A motivating example would be great to see
- A common reply to PRs without issues is that changes typically start with a proposal. (which would include a motivating example and a discussion into the design choices, rather than in the PR)
- Hidden fields are not visible across package imports, will this cause issue with your desired goals?

myitcv:
Closing because this is superseded by https://review.gerrithub.io/c/cue-lang/cue/+/532797

Issue #130, opened 2021-07-03
cmd/cue: vet: literal-only mode 


cueckoo:
_Originally opened by @rogpeppe in https://github.com/cuelang/cue/issues/130_

Cue makes for a potentially nice representation format even just as a literal representation of the data, without any of the expressive power of identifiers, functions, definitions, etc.

As suggested by @crawshaw [here](https://twitter.com/davidcrawshaw/status/1177550618178850817?s=20), it might be good to have a literal-only vet mode (`--literal` ?).

Such a subset of cue would be more amenable to parsing by other languages that don't have the full cue machinery, while still providing a very readable syntax for config files and the like.

This could also be paired with a `cue` export type to render the output in Cue rather than JSON.

cueckoo:
_Original reply by @mpvl in https://github.com/cuelang/cue/issues/130#issuecomment-536052437_

This is consistent with having a `cue export` mode type of CUE itself. That would be exactly that output.

Regarding the `vet` mode, I think the `-c` (concrete) option already does what you want. You may want to verify.

cueckoo:
_Original reply by @rogpeppe in https://github.com/cuelang/cue/issues/130#issuecomment-536305266_

> Regarding the `vet` mode, I think the `-c` (concrete) option already does what you want. You may want to verify.

That's a bit different. The `-c` option verifies that the output is concrete, but it still requires a full Cue interpreter to determine that concrete output. For example, `cue vet -c` succeeds on the following code:
```
import "strings"
x :: {
	y: strings.Join(["hello", "goodbye"], ",")
	...
}
z :: {
	z: 45
	...
}
w: x & z
```

So if you wanted to write a some code that reads the subset of cue accepted by `cue vet -c`, you'd have to implement all of cue, including the standard library. I'd like to see a format that really is just Cue literals, that could be implemented relatively straightforwardly in multiple languages and admits efficient and algorithmically simple runtime performance.

It could even have its own name - how about "Q (CUE) Strict Object Notation" or QSON for short?

cueckoo:
_Original reply by @mpvl in https://github.com/cuelang/cue/issues/130#issuecomment-536520355_

@rogpeppe Cute how the S is repurposed as strict.

[...]

PR #1434, opened 2021-12-15
encoding/cuedata: Add CUEdata package

### LINKED ISSUE #1435: Storing and querying CUE in database
At balena.io we make extensive use of manifest type files we call contracts, they are a mix of data and schema, and we have an internal database for storing and querying contracts. Contracts are stored in JSONB fields and schema is defined using JSONSchema. I opened a discussion about this see #1250.

We are keen to adopt CUE for contracts for a few reasons. Mixing data and schema with CUE is more natural and much less verbose than JSONSchema. CUE has some nice ergonomic improvements over plain JSON (see #130), like no requirement to quote property names and inline objects with a single property. Also, contracts can be evaluated together to validate compatibility, solve constraints and generate new output formats and CUE has some nice properties for evaluation vs JavaScript with JSON/JSONSchema.

As far as I know there is no way to encode and store complex CUE files include expressions and literals as data, so that concrete values can be queried and resulting files can be decoded back to CUE and expressions evaluated. 

CUE to JSONSchema is an option, but it would limit encoding and decoding to what can be expressed as JSONSchema.

I propose an (en|de)coder that can convert any arbitrarily complex CUE file to a plain object format, with concrete data as literals and CUE expressions serialized to string and appended as a special properties for later decoding. Internally I am calling this format CUEdata but I am totally open to any name, such as those mentioned in #130, QSON or CUEL.

We do still have a need for JSONSchema, contracts can also include queries that are defined using a schema, we have a JSONSchema to SQL compiler. I propose an attribute could be used to control how a struct is encoded. When a format attribute is specified; CUE expressions are encoded to that format instead of being serialized to a syntax string.

cc @myitcv @mpvl

codewithcheese:
Introduces a new (en|de)coder to covert any CUE file to and from a literal data format. Non concrete expressions in each struct are serialized into a syntax string and added as a field, labelled `$$cue`.

This allows for CUE files to be stored and transmitted as structured data. Our use case at balena is to store CUE files in a document database and query against the concrete values, the resulting documents can then be decoded back to CUE and evaluated as CUE. I think others would share a similar use case, hence I am creating this PR. If this is not acceptable, we will create it as a standalone tool.

I have named the format `CUEdata`, but I am open to alternatives. I see in issue #130 `QSON` as in "Q (CUE) Strict Object Notation" and `CUEL` as in "CUE Literal" were suggested. 

This address #1250, for our needs its a more complete solution than JSONSchema, since it allows the document to contain arbitrarily complex CUE expressions that could not be encoded as JSONSchema.

I am new to the Go language and this is an implementation by imitation, I have probably copied some techniques that the CUE code base would like to move away from. I am not aware of all the Go idioms. Please feel free to let me know any better solutions.

Resolves #1435


myitcv:
@codewithcheese thank you for raising this PR, we're delighted to welcome new contributions!

Referring to the [contribution guide](https://github.com/cue-lang/cue/blob/master/doc/contribute.md#before-contributing-code), please can I check where this particular approach/design has been previously agreed?

I can quite understand that you might have wanted to jump in and have a get familiar with the code. At the same time, we want to be respectful of people's time, and discuss changes ahead of time:

> Excluding very trivial changes, all contributions should be connected to an existing issue. Feel free to open one and discuss your plans. This process gives everyone a chance to validate the design, helps prevent duplication of effort, and ensures that the idea fits inside the goals for the language and tools. It also checks that the design is sound before code is written; the code review tool is not the place for high-level discussions.

codewithcheese:
hey @myitcv, I opened this discussion #1250 but unfortunately it was during the summer holidays and it did not receive much attention at the time. I recently discovered #130, and realized in retrospect, that I should have likewise utilized an issue to explore how to solve the use case we were facing.

myitcv:
> unfortunately it was during the summer holidays and it did not receive much attention at the time

This is a very fair point. Thank you for being patient!

We absolutely don't want to discourage contributions by referring people to the contributing guide. Instead we want to be as sensitive and respectful of people's time, as I mentioned above. 

So my suggestion would be as follows: let's either reuse #130, convert #1250 to an issue, or create a new issue (as you see fit) to discuss in more detail what you're proposing, the use case motivating it, how existing mechanisms fall short etc. The reason I stress an _issue_ here is that our workflow for proposals is issue-based, not discussions. That might seem like a small thing, but with lots of issues/discussions, we need some process. 

If we go quiet, please feel free to ping us with `@` mentions. Whilst I'm not trying to make excuses, it's always a challenge trying to balance the many demands of an open source project, and so therefore pinging people (like on the Go project when issues or CLs go quiet) is entirely reasonable and expected. 

Thank you again for your contribution, but above all else your patience and understanding during the summer months. I will help to move along whatever issue discussion so we get a quick decision on next steps.

Issue #1413, opened 2021-12-05
cue: unexpected panic in BuildInstances


rogpeppe:
commit 94db2eb171e7934e5d773d275bfe0d7558333c56

I ran the following testscript:
```
# pull in go.sum
go mod tidy
go run .
-- config/config.cue --
x: true
-- main.go --
package main

import (
	"log"

	"cuelang.org/go/cue/cuecontext"
	"cuelang.org/go/cue/load"
)

func main() {
	ctx := cuecontext.New()
	insts := load.Instances([]string{"."}, &load.Config{
		Dir: "config",
	})
	_, err := ctx.BuildInstances(insts)
	if err != nil {
		log.Fatal(err)
	}
	log.Println("ok")
}
-- go.mod --
module example.com

go 1.18

require cuelang.org/go v0.4.1-beta.6

require (
	github.com/cockroachdb/apd/v2 v2.0.1 // indirect
	github.com/emicklei/proto v1.6.15 // indirect
	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b // indirect
	github.com/google/uuid v1.2.0 // indirect
	github.com/mpvl/unique v0.0.0-20150818121801-cbe035fff7de // indirect
	github.com/pkg/errors v0.8.1 // indirect
	github.com/protocolbuffers/txtpbfmt v0.0.0-20201118171849-f6a6b3f636fc // indirect
	golang.org/x/net v0.0.0-20200226121028-0de0cce0169b // indirect
	golang.org/x/text v0.3.2 // indirect
	golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543 // indirect
	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
)
```

I see this panic:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x60c018]

goroutine 1 [running]:
cuelang.org/go/internal/core/adt.(*Vertex).getNodeContext(0xffffffffffffff01?, 0xc000345c68?)
	/home/rogpeppe/src/go/pkg/mod/cuelang.org/go@v0.4.1-beta.6/internal/core/adt/eval.go:911 +0x18
cuelang.org/go/internal/core/adt.(*OpContext).Unify(0xc0004bc340, 0x0, 0x5)
[...]

PR #1444, opened 2021-12-29
cue: fix NPE in `BuildInstances`.

mattmoor:
I hit the following NPE downstream playing with Cue for sigstore/cosign policies:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x10 pc=0x100847180]

goroutine 1 [running]:
cuelang.org/go/internal/core/adt.(*Vertex).getNodeContext(0x0, 0x140005d3d40)
        /Users/mattmoor/go/pkg/mod/cuelang.org/go@v0.4.0/internal/core/adt/eval.go:911 +0x20
cuelang.org/go/internal/core/adt.(*OpContext).Unify(0x140005d3d40, 0x0, 0x5)
        /Users/mattmoor/go/pkg/mod/cuelang.org/go@v0.4.0/internal/core/adt/eval.go:172 +0x38
cuelang.org/go/internal/core/adt.(*Vertex).Finalize(...)
        /Users/mattmoor/go/pkg/mod/cuelang.org/go@v0.4.0/internal/core/adt/composite.go:445
cuelang.org/go/cue.newVertexRoot(0x1400020c1e0, 0x140005d3d40, 0x0)
        /Users/mattmoor/go/pkg/mod/cuelang.org/go@v0.4.0/cue/types.go:615 +0x70
cuelang.org/go/cue.newValueRoot(0x1400020c1e0, 0x140005d3d40, {0x102353138, 0x0})
        /Users/mattmoor/go/pkg/mod/cuelang.org/go@v0.4.0/cue/types.go:624 +0x4c
cuelang.org/go/cue.(*Context).make(0x1400020c1e0, 0x0)
        /Users/mattmoor/go/pkg/mod/cuelang.org/go@v0.4.0/cue/context.go:246 +0x70
cuelang.org/go/cue.(*Context).BuildInstances(0x1400020c1e0, {0x14000010ae8, 0x1, 0x1})
        /Users/mattmoor/go/pkg/mod/cuelang.org/go@v0.4.0/cue/context.go:147 +0xf8
```

Examining the callstack it seems that `c.make(v)` is called when `v` is `nil`, which makes sense since the error path appends the error and then falls through.

### DIFF
--- a/cue/context.go
+++ b/cue/context.go
@@ -143,8 +143,10 @@ func (c *Context) BuildInstances(instances []*build.Instance) ([]Value, error) {
 		v, err := c.runtime().Build(nil, b)
 		if err != nil {
 			errs = errors.Append(errs, err)
+			a = append(a, c.makeError(err))
+		} else {
+			a = append(a, c.make(v))
 		}
-		a = append(a, c.make(v))
 	}
 	return a, errs
 }
--- a/cue/context_test.go
+++ b/cue/context_test.go
@@ -16,10 +16,14 @@ package cue_test
 
 import (
 	"fmt"
+	"io/ioutil"
 	"testing"
 
 	"cuelang.org/go/cue"
+	"cuelang.org/go/cue/build"
 	"cuelang.org/go/cue/cuecontext"
+	"cuelang.org/go/internal/cuetxtar"
+	"github.com/rogpeppe/go-internal/txtar"
 )
 
 func TestNewList(t *testing.T) {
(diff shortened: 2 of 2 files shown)


verdverm:
Do you have some sample code that demonstrates hitting this error, which could be made into a regression test?

mattmoor:
I actually hit this a fair number of NPEs (possibly all this), but the one I have handy is something like:
```cue
    sigs: [
        for sig in #image.sigs
        if sig & {sub: "keyless@distroless.iam.gserviceaccount.com")     // NOTE: ')' instead of '}'
        {sig}
    ]
```

mattmoor:
I have a test for this crafted, and it looks like @dlorenc code that I started from is missing this check after the `load.Instances` call:
```
	if instance.Err != nil {
		t.Fatal(instance.Err)
	}
```

verdverm:
I know this has been talked about elsewhere, but this issue brings up the idea of fuzzing. With Go 1.18 approaching, it might be worthwhile to see if https://go.dev/blog/fuzz-beta could be helpful in uncovering more like this.

dlorenc:
Friendly ping here! Any chance this could get a review?

myitcv:
Apologies for the delay in replying here, and thanks for proposing a fix @mattmoor (👋  to @dlorenc too!)

I've just commented on https://github.com/cue-lang/cue/issues/1413, which is the issue that captures the problem. 

As I said in my comment, I'm not clear that fix proposed here is necessarily correct, because I think we should be returning a `[]Value` such that `len([]Value) == len([]*BuildInstace)` (so to speak). 

But per my comment I'll defer to @mpvl. We'll get this fixed asap either way.

Thanks for your patience.

cue/context.go
```diff
@@ -143,6 +143,7 @@ func (c *Context) BuildInstances(instances []*build.Instance) ([]Value, error) {
 		v, err := c.runtime().Build(nil, b)
 		if err != nil {
 			errs = errors.Append(errs, err)
+			continue
```
mpvl: The only semantics that makes sense is to append an error value, or to, in case of any error, return no value at all. Just skipping an erroneous value like here will cause a misalignment in the values returned w.r.t. to the inputs.

Because an error Value can be useful, and to be more compatible with the deprecated `cue.Build`, I would suggest adding a Value representing the error, rather than skipping it. So add `c.makeError(err)` before continuing.

mattmoor: LMK if this is what you had in mind!

cue/context_test.go
```diff
@@ -80,3 +84,58 @@ func TestNewList(t *testing.T) {
 		})
 	}
 }
+
+func TestBuildInstancesSuccess(t *testing.T) {
```
mpvl, line 88: I personally would use a table-driven test here. But it seems in general that such tests could benefit from some common infrastructure using txtar files.

This can be done in a followup CL.

mattmoor:
Whoops just noticed DCO was busted, fixed that.

myitcv:
@mattmoor - thanks again for the patience here.

Per the [contribution guide](https://github.com/cue-lang/cue/blob/master/doc/contribute.md#good-commit-messages), please can you squash down the commits? Then we can approve and import into Gerrit for merging.

mattmoor:
should be a single commit now.

myitcv (approved):
Thanks for your patience here, @mattmoor

myitcv:
@mattmoor submitted via https://review.gerrithub.io/c/cue-lang/cue/+/532990.

Note that per the [contribution guide on commit messages](https://github.com/cue-lang/cue/blob/master/doc/contribute.md#good-commit-messages) I tweaked the commit message on the way. GitHub has a ~terrible~ non-existent interface for reviewing commit messages, which is where Gerrit wins by default. So I just note this for any future changes you might submit.

Thanks again for your patience here.

PR #1450, opened 2022-01-03
Correction of error messages in tutorials

riita10069:
The error message displayed in the actual tutorial was different, so it was modified.


myitcv:
Thanks - I'm actually fixing this as part of a wider series of fixes for the Kubernetes tutorial, but appreciate you creating the PR in the first place.

PR #1566, opened 2022-03-03
internal/core/export: avoid ".0" suffix for int53 representable numbers

brandonbloom:
When encoding Go values, it is common for data parsed from JSON to
contain float64 values when the expected information is an integer.
During export, converting numbers to ASTs attempts to preserve float vs
int encoding, but in the absence of hints, fallsback to float encoding,
producing ".0" formatted output.

This change includes an added heuristic, which fallsback to float
encoding iif the value cannot be precisely represented as an integer
within a float64 value. That is, values within the the range of a 53-bit
signed integer are printed as integers, otherwise as floats.

### DIFF
--- a/internal/core/export/export_test.go
+++ b/internal/core/export/export_test.go
@@ -214,6 +214,22 @@ func TestGenerated(t *testing.T) {
 		out: ``, // empty file
 	}, {
 		in: func(r *adt.OpContext) (adt.Expr, error) {
+			n := 1.2
+			v := ctx.Encode(n)
+			_, x := value.ToInternal(v)
+			return x, nil
+		},
+		out: `1.2`,
+	}, {
+		in: func(r *adt.OpContext) (adt.Expr, error) {
+			n := float64(3)
+			v := ctx.Encode(n)
+			_, x := value.ToInternal(v)
+			return x, nil
+		},
+		out: `3`,
+	}, {
+		in: func(r *adt.OpContext) (adt.Expr, error) {
 			v := &adt.Vertex{}
 			v.SetValue(r, adt.Finalized, &adt.StructMarker{})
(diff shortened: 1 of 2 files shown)


brandonbloom:
As is, there is a behavior change leading to a failing yaml encoding test. Input of `2.` is being encoded as `2` in yaml, but `2.` is expected. This behavior is arguably OK, since the parsed result will be valid, but it is indeed a behavioral change.

A full solution would be to introduce a third kind of number token. Presently, there is `Int` and `Float`, but there could be an untyped `Number`, which behaves similarly to untyped literals in Go. This way, the check for representability would only occur for `Number`. The test would be fixed because parsing `2.` would produce Float, and parsing `2` would produce `Int`, but converting from any Go type would produce `Number`.

This, however, is a larger change, so I'm not going to embark on it until receiving guidance from the Cue team. Thanks.

mpvl:
This really should be an issue, not a PR, as it is quite a subtle change.

I see two issues here based on your problem statement.

1. Go doesn't preserve precision when parsing JSON
2. CUE adds perhaps unnecessary .0 to JSON

The solution is not to make the representation of CUE to the AST lossy.

The solution to 1. could be to _only_ allow converted numbers to be integer when converting a `float64` that is inside an `interface{}` type.

The solution to 2. could be to allow the `.0` to be dropped even for float types when converting CUE to JSON, as JSON is inherently doesn't distinguish between int and float anyway.

brandonbloom:
> This really should be an issue, not a PR

If you prefer to have an issue for whatever tracking purposes, I've gone ahead and created one: https://github.com/cue-lang/cue/issues/1693

Will follow-up over there.

brandonbloom:
See discussion in #1693

PR #1657, opened 2022-04-22
encoding/json: unit test for json Validate

vaikas:
As part of sigstore, we use cue for policy language. I am a total cue n00b, so I'm sure I'm holding something wrong here. But the non-functioning test case here is roughly what I would like to understand how to validate properly with cue. Say I want to say that the "a.items" array must have at least 5 entries, tried to boil it down to smallest test case. the "items" and "nonitems" are dynamic, they are user generated keys.

I'd be super grateful if somebody could take a look and tell me how I could do this, or alternatively tell me that it just won't go :)

The last test case fails with this:
```
--- FAIL: TestValidate (0.00s)
    /Users/vaikas/projects/go/src/github.com/cue-lang/cue/encoding/json/json_test.go:206: Unexpected error, want: conflicting values string and 4 got: invalid operand "a" (found string, want list or struct) (and 1 more errors)

```

I was going off of this example:
https://cuelang.org/docs/tutorials/tour/expressions/conditional/

Signed-off-by: Ville Aikas <vaikas@chainguard.dev>

### DIFF
--- a/encoding/json/json_test.go
+++ b/encoding/json/json_test.go
@@ -23,8 +23,10 @@ import (
 
 	"github.com/stretchr/testify/assert"
 
+	"cuelang.org/go/cue"
 	"cuelang.org/go/cue/ast"
 	"cuelang.org/go/cue/format"
+	"cuelang.org/go/internal/core/runtime"
 )
 
 func TestExtract(t *testing.T) {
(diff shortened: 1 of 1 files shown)


mvdan:
Late to replying here - in the future, I think questions are better suited for the discussions tab or Slack, so that we don't mistake your question for a patch contribution :)

I believe what you want is https://pkg.go.dev/cuelang.org/go/pkg/list@v0.5.0-rc.1#MinItems, for example: https://tip.cuelang.org/play/?id=--Ep9vOrwU1#cue@export@cue

I'm going to close this for now, because I don't think that adding these tests to the json package helps us with code coverage.

PR #1670, opened 2022-04-26
cue: float64 support for zero

### LINKED ISSUE #1669: strconv.FormatFloat fails with zero (0.0)
master 00c5ddf5

Yes

### What did you do?

t6: strconv.FormatFloat(0.0, 102, -1, 64)

"0"

### What did you see instead?

cannot use 0.0 (type float) as float64 in argument 0 to strconv.FormatFloat: value was rounded down

kcburge:
Fix the error when attempting to use strconv.FormatFloat with a zero value:

cannot use 0.0 (type float) as float64 in argument 0 to strconv.FormatFloat: value was rounded down

The tests for maximum and minimum in Float64() do not allow for zero.  The below check incorrectly sees zero as smaller, because the Sign() of zero (0) is less than the Sign() of smallestPosFloat64 (1).  So, it was clear this logic was not accounting for zero.

		if n.X.Cmp(smallestPosFloat64) == -1 {

closes #1669 

Signed-off-by: Kevin Burge <kcburge@pm.me>

### DIFF
--- a/cue/types.go
+++ b/cue/types.go
@@ -528,6 +528,9 @@ func (v Value) Float64() (float64, error) {
 	if err != nil {
 		return 0, err
 	}
+	if n.X.IsZero() {
+		return 0.0, nil
+	}
 	if n.X.Negative {
 		if n.X.Cmp(smallestNegFloat64) == 1 {
 			return -0, ErrAbove
--- a/cue/types_test.go
+++ b/cue/types_test.go
@@ -460,6 +460,14 @@ func TestFloat(t *testing.T) {
 		fmt:     'g',
 		kind:    IntKind,
 	}, {
+		value:   "0.0",
+		float:   "0.0",
+		mant:    "0",
+		exp:     -1,
+		float64: 0.0,
+		fmt:     'g',
+		kind:    FloatKind,
+	}, {
 		value:   "1.0",
 		float:   "1.0",
 		mant:    "10",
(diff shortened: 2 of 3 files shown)


kcburge:
closes #1669

mpvl:
Good catch.

myitcv:
I will leave @mpvl to review the change itself, @kcburge, but just a comment on the DCO. The only requirement to assert the DCO is to include the trailer `Signed-off-by` in the commit message (per the [contribution guide](https://github.com/cue-lang/cue/blob/master/doc/contribute.md#asserting-a-developer-certificate-of-origin)). So one change that will be required here is to remove the DCO text itself from the commit message.

mvdan (approved):
SGTM. I agree with Marcel - nice catch!

@myitcv I'm actually not sure why the DCO check is failing. Its error says that all commits must have the header, and I see the header in the only commit in this PR. https://github.com/cue-lang/cue/pull/1709 looks the same and it's green, for example.

myitcv:
@mvdan strange. Not sure what's going on here, because it looks ok to me. Indeed `git` can also log the trailer:

```
git log --pretty='%h %an %(trailers:key=Signed-off-by)'
```

@kcburge - please can you tweak the first line of the commit message per @mpvl's change to the title of this PR?

```
cue: float64 support for zero
```

This is inline with the [contribution guide on commit messages](https://github.com/cue-lang/cue/blob/master/doc/contribute.md#good-commit-messages).

Perhaps a force-push on the fixed commit message will also fix the DCO check... we'll see.

mvdan:
@kcburge friendly ping about Paul's comment above, when you have the time :) That way we can import this change into Gerrit and merge it.

If you don't have the time to do that in the next week or two, no worries - I can always take over this change for you if you'd prefer.

curio77:
Any update on this?  Honestly, as nice as the idea/vision of CUE is, with issues like this it comes across as very flimsy and unusable in practice.  😔

mvdan:
Figured out the DCO error; the commit author had the email `kevin.burge@nice.com`, but the DCO had `kcburge@pm.me`. The author seems to use the latter for his open source commits, and the signature is what matters more as far as I can tell. We haven't head from Kevin in a while, and we want this fix to be merged, so I've gone ahead and pushed to his fork branch to rebase the change and fix up the commit author and message.

PR #1689, opened 2022-04-30
cue/load/tags: add `arch` to set of injectable system variables

### LINKED ISSUE #1692: Expose the GOARCH runtime constant to CLI users

I'd like to access information about the runtime that's executing the `cue` CLI, so I can fetch appropriate binaries for an outside-of-CUE-land purpose. I can already access the OS information via the `os` system tag, and it seems like a bit of an arbitrary choice not to be able to do the same for the only other similarly-situated runtime variable that Go makes available: `runtime.GOARCH`.

I'd like `arch` to be available as a system tag, when the `--inject-vars` flag is used.

I have implemented this as PR #1689.

I currently hackily script around the `cue` invocation and interrogate various `uname` outputs. I have an incomplete map of `uname -m` to possible GOARCH variants that I populate manually, and feed into the `cue` invocation as tags. The CLI runtime has this information available at its fingertips, and it doesn't seem like a stretch to ask it to expose the info.

**Additional context**

Go's `runtime` package makes 3 top-level constants available: the name of the tool that built the running binary, the OS, and the architecture. Of those, only the last 2 are relevant to the in-CUE consumer, and it seems arbitrary to make only 1 of them available as a system variable.

I don't *believe* there's a slippery-slope argument to counter, here. The associated PR (#1689) isn't serving as precedent to open the flood gates to a wider set of variables: it's just going some small way to realign the capabilities of `cue` CLI users with folks who are consuming CUE in Go.

jpluscplusm:
This adds to the set of system-derived variables that are provided iff
the user requests them. It expands the set from 6 to 7, introducing the
variable `arch` as an exact counterpart to the existing `os` variable.

The overhead for providing this is minimal, as the `runtime` package is
already imported and in scope due to the `os` variable.

Adding this variable does not seem to be opening the door to an
expectation that the set of system variables is expandable to meet
individual users' needs. Instead, adding this system variable merely
completes the exposure of the high-level set of variables that are
*already* available to the Go-based consumer, but are not trivially
available to the CLI-based CUE consumer. Of the 2-of-3 variables exposed
by https://pkg.go.dev/runtime#pkg-constants but *not* exposed by CUE,
only `arch` reflects the system on which the CUE CLI is being executed.

The `cue injection` help text has been updated to reflect the full set
of possible values arch values, taken from
https://go.googlesource.com/go/+/ec5170397c724a8ae440b2bc529f857c86f0e6b1/src/go/build/syslist.go#11,
which was the commit that released go1.17 - the currently required CUE
go version in go.mod. This list has been reformatted upstream since
then, but the contents (most recently as at
https://go.googlesource.com/go/+/3c6a5cdb9a29c0e3b12cdaa8ab81ba22d989e3b0/src/go/build/syslist.go#50)
are unchanged.

No tests have been added in this commit. This is for a few reasons:

- the existing `os` variable appears to have no direct tests, so a
  pattern for testing relatively hard-to-mock, machine-specific
  interfaces wasn't available to copy.
- I'm unsure how this would be testable, without the test merely
  becoming a pass-through test of what runtime.GOARCH returns.
- the not-directly-tested nature of the existing `os` variable suggests
  that testing this kind of runtime-derived variable was considered
  non-trivial, non-critical, or both.

Accepting this PR would resolve #1692.

Signed-off-by: Jonathan Matthews <github@hello.jonathanmatthews.com>

### DIFF
--- a/cmd/cue/cmd/help.go
+++ b/cmd/cue/cmd/help.go
@@ -351,6 +351,13 @@ The following variables are supported:
                 freebsd   illumos   ios       js (wasm)
                 linux     netbsd    openbsd   plan9
                 solaris   windows
+   arch       architecture identifier of the current system. Valid values:
+                386        amd64        amd64p32  arm
+                armbe      arm64        arm64be   loong64
+                mips       mipsle       mips64    mips64le
+                mips64p32  mips64p32le  ppc       ppc64
+                ppc64le    riscv        riscv64   s390
+                s390x      sparc        sparc64   wasm
    cwd        working directory
    username   current username
    hostname   current hostname
--- a/cue/load/tags.go
+++ b/cue/load/tags.go
@@ -58,6 +58,11 @@ func DefaultTagVars() map[string]TagVar {
 				return ast.NewString(runtime.GOOS), nil
 			},
 		},
+		"arch": {
+			Func: func() (ast.Expr, error) {
+				return ast.NewString(runtime.GOARCH), nil
+			},
+		},
 		"cwd": {
 			Func: func() (ast.Expr, error) {
 				return varToString(os.Getwd())


cue/load/tags.go
```diff
@@ -88,6 +88,11 @@ func DefaultTagVars() map[string]TagVar {
 				return ast.NewLit(token.INT, string(hx[:])), nil
 			},
 		},
+		"arch": {
+			Func: func() (ast.Expr, error) {
+				return ast.NewString(runtime.GOARCH), nil
+			},
+		},
```
mvdan: nit: place it next to `os`, just like in the docs?

mvdan (approved):
SGTM

cue/load/tags.go
jpluscplusm: @mvdan I deliberately played it safe, putting this at the end of the list, just in *case* someone out there is relying on the current "ordering" of the struct that `DefaultTagVars()` returns. I'm aware that this is an utter corner case, probably meaningless given the unordered nature of the return type, and no-one **should** be doing that - but that was the motivation behind not placing it next to `os`.

Thoughts?

mvdan: Maps in Go are entirely unordered, so it is literally the same no matter what order you put the keys in :)

cmd/cue/cmd/help.go
```diff
@@ -351,6 +351,13 @@ The following variables are supported:
                 freebsd   illumos   ios       js (wasm)
                 linux     netbsd    openbsd   plan9
                 solaris   windows
+   arch       architecture identifier of the current system. Valid values:
```
myitcv, line 354: Nit: please can we make the padding between columns 2 spaces, like `os`?

Also: which Go version is the reference for this list? The minimum Go version that CUE supports?

myitcv (commented):
LGTM with one nit (noting @mvdan's nit).

Happy to approve once we establish a documented answer to my question.

mvdan:
Friendly ping @jpluscplusm; we can merge as soon as the nits above are resolved :)

jpluscplusm:
Apologies for the radio silence on this PR. I'll push some commits to bring it up to scratch later ~today~ tomorrow.

cue/load/tags.go
jpluscplusm: Fixed in the most recent force-push.

cmd/cue/cmd/help.go
jpluscplusm: Formatting changed to double-spacing; the slightly longer lines don't look too bad to my eyes :-)

Commit message and PR updated to reference that the list of architectures was taken from the golang commit that released go1.17. Go1.17 was chosen as

(a) it's the oldest version that (I believe) CUE can be built with, given the version constraint found in  `go.mod` and
(b) the underlying information about supported architectures is still valid in golang's `master` branch, today.

jpluscplusm:
👋 @mvdan @myitcv PR and commit updated to resolve issues, hopefully!

mvdan:
Thank you - still looks good :) I will be importing to Gerrit shortly.

mvdan:
Now imported as https://review.gerrithub.io/c/cue-lang/cue/+/543335 and running CI. The only change made to the commit is that I added a "fixes" line at the end.

jakelogemann (approved):

PR #1704, opened 2022-05-06
cue/format/printer: fix indentation for clauses

### LINKED ISSUE #1544: cue/format: let/comprehension incorrectly indented when first expression in file

<pre>
$ cue version
cue version +a629625c linux/arm64
</pre>

Yes

### What did you do?

```
cp x.cue x.cue.golden
cp y.cue y.cue.golden
cp z.cue z.cue.golden

exec cue fmt x.cue y.cue z.cue

cmp x.cue x.cue.golden
cmp y.cue y.cue.golden
cmp z.cue z.cue.golden

-- x.cue --
if true {
	x: 5
}
-- y.cue --
let x = 5
y: x
-- z.cue --
for v in ["a", "b"] {
	(v): v
}
```

Passing test

### What did you see instead?

```
> cp x.cue x.cue.golden
> cp y.cue y.cue.golden
> cp z.cue z.cue.golden
> exec cue fmt x.cue y.cue z.cue
> cmp x.cue x.cue.golden
--- x.cue
+++ x.cue.golden
@@ -1,3 +1,3 @@
- if true {
+if true {
        x: 5
 }

FAIL: /tmp/testscript1129945673/repro.txt/script.txt:7: x.cue and x.cue.golden differ
error running repro.txt in /tmp/testscript1129945673/repro.txt
```

But note that currently all comparisons fail because in each case the expression is indented by a single space.

FogDong:
Signed-off-by: Fog Dong <wuwuglu19@gmail.com>

This PR reset the allow for clauses at the position 0.

Fixes #1544


FogDong:
PTAL😀  /cc @myitcv

myitcv:
I'll defer to @rogpeppe here :)

cue/format/testdata/indentation.input
```diff
@@ -0,0 +1,3 @@
+if true {
+  x: 5
+}
```
mvdan: nit: add the missing trailing newline

mvdan (commented):
what about adding a check in `writeWhitespace` to never allow starting a file with a whitespace character? that seems to never make sense.

an alternative is why `allowed` begins with `blank` being set when printing the first token; it seems like we should start with just `ignore`, the zero value.

FogDong:
> what about adding a check in `writeWhitespace` to never allow starting a file with a whitespace character? that seems to never make sense.
> 
> an alternative is why `allowed` begins with `blank` being set when printing the first token; it seems like we should start with just `ignore`, the zero value.

That makes more sense, thanks for the review, I'll fix it as suggested.

FogDong:
PTAL 😄 /cc @mvdan

mvdan:
@Lyoness and myself are reviewing this together :) give us a couple of days if that's okay.

mvdan:
I had a hunch that we could instead investigate why `f.allowed` had `blank` set when we were printing the first clause token. I debugged that a bit with @Lyoness and we ended up with a different fix, which you can find at https://review.gerrithub.io/c/cue-lang/cue/+/538624.

What do you think, @FogDong?

FogDong:
LGTM, you guys know cue better than me after all. @mvdan 

Still, I think it's better for the community if we can have some more discussion before another PR. Although it took some time to investigate this issue and made my first PR, it's time to close it.

mvdan:
Certainly - and the last thing I want to do is to discourage further contributions :)

I did try to nudge the discussion in another direction in an earlier comment:

> an alternative is why `allowed` begins with `blank` being set when printing the first token; it seems like we should start with just `ignore`, the zero value.

However, it was hard for me to say whether that idea would turn out to be a good fix without trying it. It's far from ideal if I ask you to implement a solution without me knowing ahead of time whether it will work or not - I'm new to the `cue/format` codebase as well.

So I thought it would be best to just try and see what happens. I'm still not sure whether Paul or Marcel will say that the second solution is better - if not, perhaps we'll come back here.

PR #1766, opened 2022-06-16
support buildinfo in version sub-command

### LINKED ISSUE #1697: cmd/cue: version information does not include VCS information for devel builds

<pre>
$ cue version
cue version +bd9ae537 linux/arm64
</pre>

Yes

### What did you do?

```
# Set gobin
mkdir gobin
env GOBIN=$WORK/gobin

# Install cue
exec git clone https://github.com/cue-lang/cue
cd cue
exec git checkout bd9ae5371cc23c54cb3cb90c0bd5a1fa4d82bc92
go install ./cmd/cue

# Check version
exec $GOBIN/cue version
stdout 'cue version +bd9ae537 linux/arm64'
```

A passing test

### What did you see instead?

```
# Check version (0.008s)
> exec $GOBIN/cue version
[stdout]
cue version (devel) linux/arm64
> stdout 'cue version +bd9ae537 linux/arm64'
FAIL: /tmp/testscript321861946/repro.txt/script.txt:13: no match for `cue version +bd9ae537 linux/arm64` found in stdout
```

cc @rogpeppe

djosephsen:
Intent of this PR is to include VCS version info as described in #1697
Fixes #1697

Signed-off-by: Dave Josephsen djosephsen@fastly.com

### DIFF
--- a/cmd/cue/cmd/testdata/script/export_err.txt
+++ b/cmd/cue/cmd/testdata/script/export_err.txt
@@ -9,6 +9,27 @@ cmp stderr expect-stderr
 ! cue export ./exporterr
 cmp stdout expect-stdout
 cmp stderr expect-stderr
+
+! cue export repeated.cue -o openapi:export.json
+! stdout .
+cmp stderr expect-stderr-openapi
+
+! cue eval repeated.cue --out openapi
+! stdout .
+cmp stderr expect-stderr-openapi
+
+-- repeated.cue --
+info: {
+    title: "test title"
+}
+
+info: {
+    title: "test title 2"
+}
+-- expect-stderr-openapi --
+info.title: conflicting values "test title 2" and "test title":
+    ./repeated.cue:2:12
+    ./repeated.cue:6:12
 -- expect-stderr --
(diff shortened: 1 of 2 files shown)


cmd/cue/cmd/version_1.18.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"golang.org/x/mod/module"
+)
+
+func runVersion(cmd *Command, args []string) error {
+	w := cmd.OutOrStdout()
+	// read in build info
+	bi, ok := debug.ReadBuildInfo()
+	if !ok {
+		// fall back to
```
mvdan: I think I started adding this line then forgot :) it should be deleted.

mvdan (approved):
Thank you!

Could you please squash the commits and write your final commit message there? This way it will be imported to Gerrit as-is, because Gerrit CLs are always a single commit.

Also, it would be nice to perhaps include:

> Note that the behavior of the command is unchanged versions of Go before 1.18,
> since the embedding of VCS information appeared in Go 1.18.

PR #1775, opened 2022-06-25
fix #1772: EncodeType doesn't unify embedded Go struct with struct definition

### LINKED ISSUE #1772: cue: EncodeType doesn't unify embedded Go struct marked with a json inline tag with struct definition

<pre>
$ cue version
cue version v0.4.3 darwin/amd64

$ go version
go version go1.17.6 darwin/amd64
</pre>

Yes

### What did you do?

When did convert a Go type to CUE value via `EncodeType`, embedded Go struct marked with a json inline tag doesn't unify with struct definition, an empty CUE field will map to CUE value of embedded Go struct.

Reproduce this via below code:
```go
package main

import (
	"fmt"

	"cuelang.org/go/cue/format"
)

func main() {
	ctx := cuecontext.New()
	val := ctx.EncodeType(&Spec{})
	node := val.Syntax()
	raw, _ := format.Node(node)
	fmt.Println(string(raw))
}

type Spec struct {
	Example Example `json:"example"`
}

type Example struct {
	Hello `json:",inline"`
	World `json:",inline"`
}

type Hello struct {
	Hello string `json:"hello"`
}

type World struct {
	World string `json:"world"`
}
```

```cue
{
	example: {
		hello: string
		world: string
	}
}
```

### What did you see instead?
```cue
{
	example: {
		"": {
			hello: string
			world: string
		}
	}
}
```

yangzhares:
Signed-off-by: Clare Yang (zhanyang) <zhanyang@cisco.com>

### DIFF
--- a/internal/core/convert/go.go
+++ b/internal/core/convert/go.go
@@ -685,16 +685,21 @@ func goTypeToValueRec(ctx *adt.OpContext, allowNullDefault bool, t reflect.Type)
 				}
 				elem = ast.NewBinExpr(token.AND, elem, v)
 			}
-			// TODO: if an identifier starts with __ (or otherwise is not a
-			// valid CUE name), make it a string and create a map to a new
-			// name for references.
 
-			// The GO JSON decoder always allows a value to be undefined.
-			d := &ast.Field{Label: ast.NewIdent(name), Value: elem}
-			if isOptional(&f) {
-				d.Optional = token.Blank.Pos()
+			if name == "" {
+				obj.Elts = append(obj.Elts, &ast.EmbedDecl{Expr: elem})
+			} else {
+				// TODO: if an identifier starts with __ (or otherwise is not a
+				// valid CUE name), make it a string and create a map to a new
+				// name for references.
+
+				// The GO JSON decoder always allows a value to be undefined.
+				d := &ast.Field{Label: ast.NewIdent(name), Value: elem}
+				if isOptional(&f) {
+					d.Optional = token.Blank.Pos()
+				}
+				obj.Elts = append(obj.Elts, d)
 			}
-			obj.Elts = append(obj.Elts, d)
 		}
 
 		// TODO: should we validate references here? Can be done using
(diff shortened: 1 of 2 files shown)


myitcv:
Thanks @yangzhares - I'll defer to @mpvl for a review here.

In the meantime, please can you take a look at the contribution guide on commit messages? 

https://github.com/cue-lang/cue/blob/master/doc/contribute.md#good-commit-messages

That gives advice on the scheme we follow there. Thanks

verdverm:
I'm curious what happens if the embedded struct does not have a json tag. A test covering this would be nice too, if it doesn't already exist somewhere

yangzhares:
@myitcv , updated commit message.

@verdverm , added two test cases, i'm not sure that is what you want

verdverm:
@yangzhares That looks to cover what I was thinking about, though now I'm wondering what happens with E.

```Go
type EmbeddedObj struct {
	B int `json:"b"`
	D string
	E string `json:"e",omitempty`
}
```

yangzhares:
Based on current logic 1) goTypeToValueRec, https://github.com/cue-lang/cue/blob/master/internal/core/convert/go.go#L693, 2) getName, https://github.com/cue-lang/cue/blob/master/internal/core/convert/go.go#L99, 3)isOptional, https://github.com/cue-lang/cue/blob/master/internal/core/convert/go.go#L139, to handle such case, especial field `E`, its CUE code should be like:
```cue
{
    b: ((int & >=-9223372036854775808) & <=9223372036854775807)
    D: string
    e: string
}
```

verdverm:
I'm not familiar with the code, mostly just thinking about test cases to cover all the scenarios, especially when code changes in the future

mvdan:
@yangzhares apologies that we hadn't gotten back to this yet. It appears that other fixes have caused conflicts here. Would you be able to fix those and rebase? I'm happy to review afterwards. Thank you!

mvdan:
I'm going to close this PR for now as I didn't get a response above, and the patch is years out of date at this point.

I'm more than happy to review another PR if someone wants to revive the patch above. Alternatively, if someone can file an issue describing the problem with an example, we can also take a look at a fix ourselves.

PR #1784, opened 2022-07-02
spec.md: Fix missing `Slice` and `si_lit` typo

addreas:
Just something  I ran into while playing around with creating a parser based  on the spec.

Another thing I noticed is  that `decimal_lit` and `decimals` look very much equivalent, or am I missing something?


mvdan:
Thanks for the patch! It looks like @myitcv already sent these fixes as part of https://review.gerrithub.io/c/cue-lang/cue/+/540569, if you would like to take a look. He removed the slicing part as the fix, because that feature does not exist right now.

> Another thing I noticed is that `decimal_lit` and `decimals` look very much equivalent, or am I missing something?

One difference that I can see is that `decimals` allows `0_1` while `decimal_lit` does not. I can't say for sure whether that distinction is correct, though, as I can't find it being mentioned anywhere else in the spec. Perhaps @myitcv or @mpvl know more.

addreas:
Oh, didn't realize. Nothing  to see here then.

Thanks for the feedback.

PR #1787, opened 2022-07-11
pkg/tool/exec: add .Default() to struct env value

addreas:
Encountered an error that was quite confusing: `invalid environment variable value "perfectly-reasonable-string"`, looking into what caused that I ended up with this little diff.

### DIFF
--- a/pkg/tool/exec/exec.go
+++ b/pkg/tool/exec/exec.go
@@ -154,7 +154,7 @@ func mkCommand(ctx *task.Context) (c *exec.Cmd, doc string, err error) {
 	// Struct case.
 	for iter, _ := ctx.Obj.Lookup("env").Fields(); iter.Next(); {
 		label := iter.Label()
-		v := iter.Value()
+		v, _ := iter.Value().Default()
 		var str string
 		switch v.Kind() {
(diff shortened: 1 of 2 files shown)


mvdan:
Thank you for contributing, and apologies that we hadn't replied yet :)

Per the [contributing docs](https://github.com/cue-lang/cue/blob/master/doc/contribute.md), if this is meant to fix a bug, please raise an issue following the bug template first. That way we can be sure that we're on the same page and can reproduce the issue before we consider a fix.

addreas:
I've left the code that required this behind so I would be more likely to close this PR than create a bug report. The test change should demonstrate the motivation for this, though.

mvdan:
Thanks - now imported to Gerrit as https://review.gerrithub.io/c/cue-lang/cue/+/1174128.

mvdan:
It turns out that an issue was filed for this half a year ago (https://github.com/cue-lang/cue/issues/2572) and https://github.com/cue-lang/cue/pull/2764 was just sent as an alternative PR, which also fixes this for lists and has more tests, so I'll go with that one assuming it's okay with you. Thank you for contributing nonetheless!

PR #1796, opened 2022-07-19
Fix the error message on "module root not defined"

sanposhiho:
This PR is to fix the following error message. It would be better to add %s to show display path.

```
module root not defined%!(EXTRA string=./basic/user)
```


mvdan:
Thanks for the PR! Is there a reason you closed it?

sanposhiho:
Just wanted to recreate the PR :)
https://github.com/cue-lang/cue/pull/1797

PR #1797, opened 2022-07-19
cue/load: Fix the error message on "module root not defined"

sanposhiho:
Signed-off-by: Kensei Nakada <handbomusic@gmail.com>

This PR is to fix the following error message. It would be better to add %s to show display path.

```
module root not defined%!(EXTRA string=./basic/user)
```


mvdan:
Thanks for the PR! Luckily, @rogpeppe came up with a clever way of getting `go vet` to spot these errors statically, so they were all spotted and fixed as part of https://review.gerrithub.io/c/cue-lang/cue/+/551766. We shouldn't have any more of these bugs, hopefully.

PR #1807, opened 2022-07-22
fix empty yaml unmarshal

### LINKED ISSUE #1790: encoding/yaml: Panic when unmarshalling YAML content that is only comments

<pre>
$ cue version
cue version v0.4.3 darwin/arm64
</pre>

Also tested on `cue version v0.4.4-0.20220706163258-2795dcc4db87`

Yes.

### What did you do?

Attempt to unmarshal YAML content that is just a comment using `encoding/yaml`.

Given `test.cue` with:

```
import "encoding/yaml"

out: yaml.Unmarshal("#test")
```

Running `cue eval test.cue`

Certainly not a panic. Likely an empty output or bottom?

### What did you see instead?

A panic:

```
panic: assertion failed: nil return value: unspecified error [recovered]
	panic: assertion failed: nil return value: unspecified error

goroutine 1 [running]:
cuelang.org/go/cmd/cue/cmd.recoverError(0x140005bfe70)
	cuelang.org/go/cmd/cue/cmd/root.go:229 +0x98
panic({0x1035261c0, 0x1400071c860})
	runtime/panic.go:838 +0x204
cuelang.org/go/internal/core/adt.Assertf(...)
	cuelang.org/go/internal/core/adt/context.go:93
cuelang.org/go/internal/core/adt.(*OpContext).evaluateRec(0x14000719ad0, 0x1400011f7c0, {0x103610a68, 0x14000113140}, 0x78?)
	cuelang.org/go/internal/core/adt/context.go:582 +0x280
...
```

critterjohnson:
Fixes #1790. This assumes that any time an evaluator returns `nil`, but no error, it is simply evaluating an empty value and returns a Bottom.

### DIFF
--- a/internal/core/adt/context.go
+++ b/internal/core/adt/context.go
@@ -583,13 +583,12 @@ func (c *OpContext) evaluateRec(env *Environment, x Expr, state VertexStatus) Va
 
 	val := c.evalState(x, state)
 	if val == nil {
-		// Be defensive: this never happens, but just in case.
-		Assertf(false, "nil return value: unspecified error")
 		val = &Bottom{
 			Code: IncompleteError,
-			Err:  c.Newf("UNANTICIPATED ERROR"),
+			Err:  c.Newf("evaluated expression has no value"),
 		}
 	}
+
 	_ = c.PopState(s)
 
 	return val


mvdan:
I think this one is subtle enough that we want @mpvl's eyes on it :)

mvdan:
We looked at this one with @mpvl; we definitely agree that the bug is valid, but we think the fix probably belongs in the encoding package rather than the evaluator. It's also not a trivial fix, because the YAML spec doesn't clearly define whether the empty input should be invalid, `null`, or something else. I'll take a deeper look when I can.

uhthomas:
I don't think there's a good answer for this. The YAML spec is very vague and implementations don't even seem to have agreement.

https://github.com/nodeca/js-yaml/issues/565

Some of the YAML spec seems to suggest that a completely empty document, or even a document just consisting of comments should not be represented at all.

https://yaml.org/spec/1.2.2/#rule-l-bare-document

Then, an explicit document denoted with `---` should be represented as null.

https://yaml.org/spec/1.2.2/#915-directives-documents

---

I guess based on that, a bottom (`_|_`) value may be appropriate? I guess it then begs the question of what `{}` would mean and whether it would better?

mvdan:
With Marcel and Roger we agreed to decode empty YAML inputs as `null` - the YAML spec seems to fairly strongly hint in that direction, and we're all happy with the resulting behavior. Thanks again for the help!

mvdan:
(for more details, see the commit message above)

PR #2028, opened 2022-10-21
Bump golang.org/x/text to a newer version

joaopapereira:
According to a security scan using [trivy](https://github.com/aquasecurity/trivy) cue was signaled as using a version of golang.org/x/text that contains the [CVE CVE-2022-32149](https://nvd.nist.gov/vuln/detail/CVE-2022-32149). Although cue doesn't use the part of the library containing this CVE, this bump will save any false positive scans done in on this tool.


myitcv (approved):
LGTM. 

Will import into Gerrit now.

PR #2049, opened 2022-10-27
feat: dependabot workflow automation for updating dependency

rajpratik71:
Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  # Dependencies listed in go.mod
+  - package-ecosystem: "gomod"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
+
+  # Dependencies listed in .github/workflows/*.yml
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


rajpratik71:
This workflow will help in updating dependencies , which will further reduce Vulnerability automatically.

image

mvdan:
Are we actually affected by either of those CVEs? I think the analysis works on a per-package or per-module basis, so it's probably picking up some false positives. It's always possible to update them as needed, like in https://github.com/cue-lang/cue/pull/2028.

I personally would prefer to not use dependabot, because it will likely attempt to update all dependencies regularly. That's not something we want to do, in general. Some dependencies need to be updated very carefully, as they could break backwards compatibility for some CUE users. Some other dependencies are rather unimportant, so keeping them very up to date isn't necessary. We also want to avoid updating dependencies unnecessarily while we're working towards a stable release, for example.

All that to say: what do you think would improve with dependabot? If you simply want us to update libraries when CVEs are reported, I'm sure we can do that without auto-updating all dependencies.

mvdan:
cc @myitcv @mpvl

myitcv:
I'm also not totally convinced of the benefits of this change. At a minimum I'd suggest we move discussion on this to an issue first (per the [contribution guide](https://github.com/cue-lang/cue/blob/master/doc/contribute.md#open-an-issue-for-any-new-problem)).

mvdan:
Closing per the above for now :) Happy to further discuss in a GitHub Discussion if you open one. Thanks!

PR #2187, opened 2022-12-14
internal/core/convert: split the cue tag at first comma after last quote

### LINKED ISSUE #1887: internal/core/convert: cue tag can't be parsed when cue contains a comma (using cuego)

<pre>
$ cue version
cue version v0.4.3 darwin/arm64
</pre>

Yes

### What did you do?

```go
package main

import (
	"fmt"

	"cuelang.org/go/cuego"
)

type Comma struct {
	Letters string `json:"letters,omitempty" cue:"=~\"^[a-z](,[a-z])*$\",opt"`
}
type Semikolon struct {
	Letters string `json:"letters,omitempty" cue:"=~\"^[a-z](;[a-z])*$\",opt"`
}

func main() {
	s := Semikolon{"a;b;c;d"}
	if err := cuego.Validate(s); err != nil {
		fmt.Println(err)
	} else {
		fmt.Println(s.Letters)
	}
	c := Comma{"a,b,c,d"}
	if err := cuego.Validate(c); err != nil {
		fmt.Println(err)
	} else {
		fmt.Println(c.Letters)
	}
}
```

```
a;b;c;d
a,b,c,d
```

### What did you see instead?

```
a;b;c;d
invalid tag "=~\"^[a-z](" for field "letters": string literal not terminated (and 1 more errors)
```

The [functions parsing the cue tag](https://github.com/cue-lang/cue/blob/2fea90a631b14946ca0ffac025e64da2d09d3e27/internal/core/convert/go.go#L78) strips everything after the first comma to allow adding options (like "opt" or "req").

I see three possible ways to change the parser:
1. strip "known" options from the end (currently "opt" and "req")
1. find and skip pairs of (not escaped) quotes from the start
1. assume the options can't contain a quote: split at the first comma after the last quote (if any)

I'm in favor of option #3 as it would be relatively simple to implement while still being generic.
I've [implemented a fix](https://github.com/cue-lang/cue/compare/master...ghostwheel42:cue:update_cue_tag_parser) and can prepare a PR if you'd like me to.

ghostwheel42:
This fixes a bug when splitting a golang struct tag containing commas.

There are 3 possible ways I can think of to fix this:
1. strip "known" options from the end (currently "opt" and "req") => fragile, we can't know all tags which could be used.
2. find and skip pairs of (not escaped) quotes from the start => complex and not really necessary, see option 3
3. assuming the options won't contain a quote: split at the first comma after the last quote (if any)

This PR implements option 3.

Fixes #1887

Signed-off-by: Alexander Graf <ghostwheel42@users.noreply.github.com>

### DIFF
--- a/internal/core/convert/go.go
+++ b/internal/core/convert/go.go
@@ -94,7 +94,7 @@ func parseTag(ctx *adt.OpContext, obj *ast.StructLit, field, tag string) ast.Exp
 
 // splitTag splits a cue tag into cue and options.
 func splitTag(tag string) (cue string, options string) {
-	q := strings.LastIndex(tag, "\"")
+	q := strings.LastIndexByte(tag, '"')
 	if c := strings.Index(tag[q+1:], ","); c >= 0 {
 		return tag[:q+1+c], tag[q+1+c+1:]
 	}
--- a/internal/core/convert/go.go
+++ b/internal/core/convert/go.go
@@ -96,7 +96,7 @@ func parseTag(ctx *adt.OpContext, obj *ast.StructLit, field, tag string) ast.Exp
 func splitTag(tag string) (cue string, options string) {
 	q := strings.LastIndex(tag, "\"")
 	if c := strings.Index(tag[q+1:], ","); c >= 0 {
-		return tag[:c], tag[c+1:]
+		return tag[:q+1+c], tag[q+1+c+1:]
 	}
 	return tag, ""
 }
--- a/internal/core/convert/go.go
+++ b/internal/core/convert/go.go
@@ -95,7 +95,7 @@ func parseTag(ctx *adt.OpContext, obj *ast.StructLit, field, tag string) ast.Exp
 // splitTag splits a cue tag into cue and options.
 func splitTag(tag string) (cue string, options string) {
 	q := strings.LastIndexByte(tag, '"')
-	if c := strings.Index(tag[q+1:], ","); c >= 0 {
+	if c := strings.IndexByte(tag[q+1:], ','); c >= 0 {
 		return tag[:q+1+c], tag[q+1+c+1:]
 	}
 	return tag, ""
@@ -112,7 +112,7 @@ func getName(f *reflect.StructField) string {
 	}
 	for _, s := range tagsWithNames {
 		if tag, ok := f.Tag.Lookup(s); ok {
-			if p := strings.Index(tag, ","); p >= 0 {
+			if p := strings.IndexByte(tag, ','); p >= 0 {
 				tag = tag[:p]
 			}
 			if tag != "" {
--- a/internal/core/convert/go.go
+++ b/internal/core/convert/go.go
@@ -78,9 +78,7 @@ func compileExpr(ctx *adt.OpContext, expr ast.Expr) adt.Value {
 
 // parseTag parses a CUE expression from a cue tag.
 func parseTag(ctx *adt.OpContext, obj *ast.StructLit, field, tag string) ast.Expr {
-	if p := strings.Index(tag, ","); p >= 0 {
-		tag = tag[:p]
-	}
+	tag, _ = splitTag(tag)
 	if tag == "" {
 		return topSentinel
 	}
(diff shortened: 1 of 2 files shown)


mvdan:
You didn't mean to link to issue number 3, right?

mvdan:
This needs tests as well :) I think the two test cases you added in the original issue would be a good start.

ghostwheel42:
I've removed the link to issue 3.
I can add some test cases, too.

ghostwheel42:
> This needs tests as well :) I think the two test cases you added in the original issue would be a good start.

I've added a test case now and found a bug (copied an old version, duh - but this is what tests are for)

I wanted to test `splitTag()` directly, but when using separate test packages (`package convert_test`) it would need to be exported only to be tested.

internal/core/convert/go.go
```diff
@@ -92,6 +90,15 @@ func parseTag(ctx *adt.OpContext, obj *ast.StructLit, field, tag string) ast.Exp
 	return expr
 }
 
+// splitTag splits a cue tag into cue and options.
+func splitTag(tag string) (cue string, options string) {
+	q := strings.LastIndex(tag, "\"")
```
mvdan:
```suggestion
	q := strings.LastIndexByte(tag, '"')
```

mvdan (approved):
Thanks! Just one minor nit. Then happy to import into Gerrit and merge.

internal/core/convert/go.go
ghostwheel42: Why don't we also use IndexByte in line 96?
Isn't a "tag" unicode-capable and we shouldn't use `strings.*Byte` but `strings.*Rune`?

mvdan: yes, you can use it there too :)

the "byte" aspect is only for the second argument, it still supports any utf8 string or valid rune in the first argument.

mvdan: to be clear, I only suggest the Byte forms because they're a bit simpler and faster. funcs like Index already fall back to IndexByte when the second argument is a one-byte string, but we can do that shortcut ourselves.

ghostwheel42: Are you sure @mvdan? What about this?
https://go.dev/play/p/Em3PNiHllR5
Hmm, okay, maybe not a good example.

ghostwheel42: I think it's okay to use IndexByte when looking for something < 0x80.
Otherwise it will fail: https://go.dev/play/p/dF8yOvB8Yjd
I could use IndexByte when looking for a comma or quotes, as utf-8 does not use bytes < 0x80 - so false positives are not expected.

mvdan: Yes, I'm pretty sure - using IndexByte or LastIndexByte is fine when your _second_ argument is plain ASCII. Your examples above would need to use the Rune forms; they break with Byte because the rune literal gets converted to byte, so information is lost.

mvdan:
The DCO bot isn't happy; it seems that one of the commits isn't signed off :) could you please fix that? For example, by squashing all the commits and ensuring the resulting commit has the trailer.

internal/core/convert/go.go
ghostwheel42: Yeah, that's what I meant by "< 0x80" - so if the byte is ascii only it is fine, if not it breaks.
This is thy I used Index() and not IndexByte() in the first place. But it's okay as long as we don't start using extended ascii charactes as delimiter here ;)

ghostwheel42:
> 



> The DCO bot isn't happy; it seems that one of the commits isn't signed off :) could you please fix that? For example, by squashing all the commits and ensuring the resulting commit has the trailer.

I think it was your "suggestion" I merged in the github web interface. I've signed all commits and rebased to master.
Should be fine now.

internal/core/convert/go.go
mvdan: I doubt we'll ever use anything other than ASCII for these separators, and if we started doing that and IndexByte broke, our tests should catch it :) if we're making syntax changes without tests, that's our fault for sure!

mvdan:
Thanks! Will be importing this into Gerrit for a merge soon.

mvdan:
> I think it was your "suggestion" I merged in the github web interface. I've signed all commits and rebased to master.

Ah, interesting, that makes sense. I'll probably avoid suggesting changes in the future.

Issue #2046, opened 2022-10-26
cmd/cue: nil pointer reference panic when using "get go"


LeonardAukea:
### What version of CUE are you using (`cue version`)?

<pre>
$ cue version
cue version v0.5.0-alpha.1

       -compiler gc
        -ldflags -s -w
     CGO_ENABLED 0
          GOARCH amd64
            GOOS darwin
         GOAMD64 v1
</pre>

### What did you do?

I ran the following commands: 

```bash
go get istio.io/api/networking/v1alpha3
go: downloading istio.io/api v0.0.0-20221025180617-53adcb30720d
go: downloading google.golang.org/protobuf v1.28.1
go: downloading google.golang.org/genproto v0.0.0-20221018160656-63c7b68cfc55
go: upgraded google.golang.org/genproto v0.0.0-20220624142145-8cd45d7dbd1f => v0.0.0-20221018160656-63c7b68cfc55
go: added istio.io/api v0.0.0-20221025180617-53adcb30720d

cue get go istio.io/api/networking/v1alpha3
```

Which returned: 

### What did you expect to see?

Some confirmation from the cli that the `cue get go` loaded the definitions successfully, or at least not return an error. 

### What did you see instead?

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x16c5ed9]

goroutine 1 [running]:
cuelang.org/go/cmd/cue/cmd.recoverError(0xc008e07e90)
        /home/runner/go/pkg/mod/cuelang.org/go@v0.5.0-alpha.1/cmd/cue/cmd/root.go:232 +0x7b
panic({0x1768dc0, 0x1d9dc70})
        /opt/hostedtoolcache/go/1.18.1/x64/src/runtime/panic.go:838 +0x207
cuelang.org/go/cmd/cue/cmd.(*extractor).extractPkg(0xc00a2ac3f0, {0xc00003a0c4, 0x2a}, 0x0)
        /home/runner/go/pkg/mod/cuelang.org/go@v0.5.0-alpha.1/cmd/cue/cmd/get_go.go:438 +0x79
cuelang.org/go/cmd/cue/cmd.(*extractor).extractPkg(0xc00a2ac3f0, {0xc00003a0c4, 0x2a}, 0xc000565380)
[...]

PR #2255, opened 2023-02-08
return error when pkg is nil

richardalberto:
Returning an error here solves the nil pointer exception detailed in #2046 

It would be good to also explain why this happens.


mvdan:
I agree that this is likely to avoid the panic, but I don't think this is a proper fix for the underlying bug at hand. I'm going to leave this open for the time being, because we might want to merge this while we figure out a better solution, but I also don't think a nil check should mark the issue as fully fixed.

Either way, any further investigation can and should happen in the issue thread, so I'll continue there :)

mvdan:
We finally merged @uhthomas's fix at https://review.gerrithub.io/c/cue-lang/cue/+/1173100, so with the original issue fixed, I think we can close this PR :) Thanks!

PR #2269, opened 2023-02-17
cmd/cue: support stringable types as map keys

addreas:
Ran into a case where [metacontroller/metacontroller](https://github.com/metacontroller/metacontroller/blob/13b175b74cc17a4cdbfbce09e9efa7342f73b7dd/pkg/controller/common/api/v1/common.go#L24) had a map where the key was a struct that implemented `encoding.TextMarshaler` and `encoding.TextUnmarshaler`. After some confusion because it was silently dropped (until i ran `cue get go --verbose`), i had a look and a fix seemed not too far off.

Not super excited about this patch, as it's a bit quick and dirty, but it did seem to do the trick.

### DIFF
--- a/cmd/cue/cmd/get_go.go
+++ b/cmd/cue/cmd/get_go.go
@@ -962,7 +962,14 @@ func supportedType(stack []types.Type, t types.Type) (ok bool) {
 	case *types.Array:
 		return supportedType(stack, x.Elem())
 	case *types.Map:
-		if b, ok := x.Key().Underlying().(*types.Basic); !ok || b.Kind() != types.String {
+		typ := x.Key()
+		if b, ok := typ.Underlying().(*types.Basic); !ok || b.Kind() != types.String {
+			ptr := types.NewPointer(typ)
+			for _, iface := range toString {
+				if types.Implements(typ, iface) || types.Implements(ptr, iface) {
+					return supportedType(stack, x.Elem())
+				}
+			}
 			return false
 		}
 		return supportedType(stack, x.Elem())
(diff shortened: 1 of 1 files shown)


mvdan:
Apologies that we didn't reply to this PR quickly. Do you have an example or reproducer for this case? In particular, we would need a form of regression test to go with the change, to make sure that it works as expected and isn't broken in the future.

addreas:
Not sure how to create a txtar for a `cue get go` command, and I've almost completely forgot about this whole thing as well. But the problem was that there existed [a type](https://github.com/metacontroller/metacontroller/blob/13b175b74cc17a4cdbfbce09e9efa7342f73b7dd/pkg/controller/common/api/v1/common.go#L24) that I wanted to `cue get go` on, but it was defined as `type RelativeObjectMap map[GroupVersionKind]map[string]*unstructured.Unstructured` and `GroupVersionKind` is a struct that implements `encoding.TextMarshaler` and `encoding.TextUnmarshaler`.

The minimal case should look something like this, I guess:
```go
package test

type Convoluted map[Key]string

type Key struct {
	ActualKey string
}

// MarshalText is implementation of  encoding.TextMarshaler
func (k Key) MarshalText() ([]byte, error) {
	return []byte(k.ActualKey), nil
}

// UnmarshalText is implementation of encoding.TextUnmarshaler
func (k *Key) UnmarshalText(text []byte) error {
	*k = Key{
		ActualKey: string(text),
	}
	return nil
}
```

Happy to close this unless you think it's a feature worth the effort.

mvdan:
Understood, thank you.

Recently we have moved towards generating Go from CUE rather than the other way around, because Go as a type system is rather limited in what it can represent. So while `cue get go` still works and it can be the correct solution in some cases, in many others I suspect the users should use `cue exp gengotypes` instead. The source schemas don't have to be CUE directly; they could be imported from JSON Schema or OpenAPI.

Given the above, and that you no longer seem to need this patch or remember the precise details, I'm inclined to close the PR for now. I'm happy to review another PR in the future if it comes with a test - or an issue describing the specific use case with an example.

Issue #2042, opened 2022-10-25
cue/eval: judgment on empty array is weird


FogDong:
### What version of CUE are you using (`cue version`)?

<pre>
$ cue version
v0.4.3
</pre>

### Does this issue reproduce with the latest release?
yes

### What did you do?

CUE Playground link: https://cuelang.org/play/?id=JpAg8kw82n3#cue@export@cue

It looks good when comparing two empty arrays:

```
emptyArr: []

if emptyArr == [] {
  hello: 1
}
```

Output: 

```
emptyArr: []
hello: 1
```

However, become strange when comparing a non-empty array to `[]`:

```
notEmptyArr: [1,2,3,]

if notEmptyArr != [] {
  hello2: 2
}
```

Output:
```
notEmptyArr: [1,2,3,]
```

### What did you expect to see?

```
notEmptyArr: [1,2,3,]
hello2: 2
```

antong:
There is a bug, that makes `list1 != list2` return false if `len(list1) != len(list2)`. It should return true:
https://github.com/cue-lang/cue/blob/dd1169c686c9e45358fa87fcdb9530fe3f6a799f/internal/core/adt/binop.go#L109-L114

I could try to fix this. But, the language specification says that lists aren't comparable: https://cuelang.org/docs/references/spec/#comparison-operators

@mpvl , should lists be comparable?

antong:
Oh, now saw #2024.

FogDong:
This one's priority is **low** since there's a **workaround** to use `if len(arr) == 0`.
This is a **user-facing** issue because in our workflow, users can write an if condition(which will be evaluated as cue value) to decide whether a step should be executed. A detailed explanation of the usage is [here](https://github.com/cue-lang/cue/issues/2232#issuecomment-1423491944).
No **time constraint**.

mvdan:
@antong is right that the implementation currently does not agree with the spec. @rogpeppe filed this as https://github.com/cue-lang/cue/issues/2024 already, thankfully.

Regardless of that discrepancy, I think we can all agree that the current logic has a tiny mistake that should be fixed. https://github.com/cue-lang/cue/pull/2290 already exists as the fix that @antong also spotted, so I'll look at getting that merged.

PR #2290, opened 2023-03-03
Fix for issue 2289 - list not equal op.

mmlt:
Before fix:
a: [] != [1]
a: false
After fix:
a: true

### DIFF
--- a/cue/testdata/lists/021_list_equality.txtar
+++ b/cue/testdata/lists/021_list_equality.txtar
@@ -99,18 +99,18 @@ feq8:  false
 feq9:  false
 feq10: false
 feq11: false
-fne0:  false
-fne1:  false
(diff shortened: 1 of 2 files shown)


mvdan (approved):
Thanks! This fix looks right.

Could you please add the DCO trailer? You can either do this via `git commit --amend -s`, or following https://github.com/cue-lang/cue/wiki/Contribution-tips-and-tricks.

mvdan:
The fix here is very small, and it's the same fix that @antong spotted originally in https://github.com/cue-lang/cue/issues/2042#issuecomment-1290878454. We want the fix to ship in v0.6, so I'm going to send it to Gerrit directly since we don't have a DCO trailer here. Thanks again for your contribution!

PR #2352, opened 2023-04-19
Add configurable stop condition for recursive schemas

spinillos:
`Openapi` doesn't have any stop condition for complex recursive schemas and generates infinite loops.

This PR adds a dummy way to add a stop condition. Mainly it adds a configurable value that indicates the maximum depth of cycle iteration. When it reaches the maximum value, it returns and prints the current result of the schema.

This new approach only works when `MaxCycleDepth` has a value bigger than 0.


spinillos:
Sorry! I opened this by mistake in this repository instead of the fork 😅.

PR #2414, opened 2023-05-22
internal/ci: remove cue-unity/unity from codereview.cfg

mvdan:
(see commit message)


mvdan:
I keep going for my "open PR" script, sorry.

myitcv:
> I keep going for my "open PR" script, sorry.

You could teach it to error if there is a `codereview.cfg` file at the git root.

mvdan:
I actually did that for a couple of days, but then ran into trouble since unity-private still has that file carried over from unity :) I didn't think it was worth the noise to remove it from internal/ci only to get it back a few days later. I'll re-apply the sanity check when we move unity-private to gerrithub.

Issue #2284, opened 2023-02-26
cue/load: hangs when importing a package with the same name as the current package


uhthomas:
### What version of CUE are you using (`cue version`)?

Installed with [pacman](https://archlinux.org/packages/community/x86_64/cue/), current version `0.4.3-1`.

There's an open issue to resolve this: https://bugs.archlinux.org/task/77648

<pre>
$ cue version
cue version (devel) linux/amd64
</pre>

### Does this issue reproduce with the latest stable release?

Yes.

### What did you do?

```
-- cue.mod/module.cue --
module: "example.com"

-- a/a/a.cue --
package a

some_field: "some value"

-- a/a.cue --
package a

import "example.com/a/a"

some_other_field: a.some_field
```

Then run `cue export`.

### What did you expect to see?

Should export successfully.

### What did you see instead?

Infinitely hangs.

mvdan:
Potentially related to https://github.com/cue-lang/cue/issues/849. Likely not an exact duplicate, but both are infinite cycles when importing packages.

bozaro:
Looks like package contains all files with same package name in all ascending directories in this module (don't ask me why): https://github.com/cue-lang/cue/blob/adbe80e41254ff45dfb3194f930db9898e1a7a47/cue/load/import.go#L138-L178
For example, package `example.com/a/b/c/a` will contains all files with `package a` from directories `./a/b/c/a`, `./a/b/c`, `./a/b`, `./a` and `./`.

In this case in packge `example.com/a/a` we got files `a/a/a.cue` and `a/a.cue`. File `a/a.cue` try import itself `import "example.com/a/a"`.

mvdan:
> Looks like package contains all files with same package name in all ascending directories in this module (don't ask me why):

Correct, that's explained in https://cuelang.org/docs/concepts/packages/#instances.

The fact that this example makes `cue/load` hang up is certainly a bug. I can't really convince myself what the right behavior would be, though.

In the original example, one can see the package `a/a` as having an implicit dependency on `a`, since `a/a` inherits the constraints from `a`, per the documentation I linked to. Then, `a` importing `a/a` seems to be what is causing an import cycle that we don't catch, causing the infinite loop.

One option would be to catch this like we already do with import cycles, resulting in an error. This is probably the simplest solution, but I'm not sure if it would break any reasonable use case.

mvdan:
> I'm not sure if it would break any reasonable use case.

[...]

PR #2435, opened 2023-06-08
cue/load: fix infinity loop on package import cycle

### LINKED ISSUE #849: cue/load: package cycles are not detected
_Originally opened by @grouville in https://github.com/cuelang/cue/issues/849_

<pre>
$ cue version
cue version 0.3.0-beta.7 darwin/amd64 (installed with brew)
</pre>

Yes

### What did you do?

It seems that `cue eval` / `cue export` goes into an infinite loop when evaluating/exporting packages with cyclic inclusion

Here is a tree representation of a little implementation, to reproduce the error :
```bash
➜  test_infinite tree
.
├── cue.mod
│   ├── module.cue
│   ├── pkg
│   │   └── top.dev
│   │       ├── la
│   │       │   └── toto.cue
│   │       └── lo
│   │           └── tata.cue
│   └── usr
└── main.cue
```
---------
*main.cue* 
```cue
package main

import (
	"top.dev/lo"
)

test: {
    value: #JO & {
        test2: "top"
    }
}
```
---------
*module.cue* 
```cue
module: "main"
```
---------
*toto.cue*
```cue
package la

import (
	"top.dev/lo"
)

#LA: {
	test?: string
}
```
---------
*tata.cue*
```cue
package lo

import (
	"top.dev/la"
)

#JO: {
	test2?: string
}
```
----------
----------

*output*
```bash
➜  test time cue export
^C
cue export  325,33s user 8,33s system 173%!c(MISSING)pu 3:12,37 total
```

I'm not sure this is an issue. I think it is, but I might be wrong :
- With a broken field in the main (other test, same structure. Change the main.cue like that :
```cue
package main

import (
	"top.dev/lo"
)

test: {
    value: #JO & {
        test2: 1
    }
}
```
=> With this main, a validation error (test2 expects a string, and I provide an int)

- With the first main showed, a warning, or maybe not an infinite inclusion [I'm not sure if this is an expected behavior]

### What did you see instead?

An infinite loop

bozaro:
Fixes #849

### DIFF
--- a/cue/build/context.go
+++ b/cue/build/context.go
@@ -66,6 +66,9 @@ func (inst *Instance) Complete() error {
 	if err != nil {
 		inst.ReportError(err)
 	}
+	for _, imp := range inst.Imports {
+		_ = imp.Complete()
+	}
 	if inst.Err != nil {
 		inst.Incomplete = true
 		return inst.Err
--- a/cue/load/import.go
+++ b/cue/load/import.go
@@ -195,7 +195,6 @@ func (l *loader) importPkg(pos token.Pos, p *build.Instance) []*build.Instance {
 		}
 
 		l.addFiles(cfg.ModuleRoot, p)
-		_ = p.Complete()
 	}
 	sort.Slice(all, func(i, j int) bool {
 		return all[i].Dir < all[j].Dir
--- a/cue/load/instances.go
+++ b/cue/load/instances.go
@@ -101,6 +101,10 @@ func Instances(args []string, c *Config) []*build.Instance {
 	}
 
 	for _, p := range a {
+		_ = p.Complete()
+	}
+
+	for _, p := range a {
 		tags, err := findTags(p)
 		if err != nil {
 			p.ReportError(err)
--- a/cue/load/loader_test.go
+++ b/cue/load/loader_test.go
@@ -103,18 +103,26 @@ files:
 imports:
     mod.test/test/sub: $CWD/testdata/testmod/sub/sub.cue`,
 	}, {
-		// TODO:
-		// - path incorrect, should be mod.test/test/other:main.
 		cfg:  dirCfg,
 		args: args("./other/..."),
 		want: `
 err:    import failed: relative import paths not allowed ("./file"):
     $CWD/testdata/testmod/other/main.cue:6:2
-path:   ""
+path:   mod.test/test/other:main
 module: mod.test/test
 root:   $CWD/testdata/testmod
-dir:    ""
-display:""`,
+dir:    $CWD/testdata/testmod/other
+display:./other
+files:
+    $CWD/testdata/testmod/other/main.cue
+
+path:   mod.test/test/other/file
+module: mod.test/test
+root:   $CWD/testdata/testmod
+dir:    $CWD/testdata/testmod/other/file
+display:./other/file
+files:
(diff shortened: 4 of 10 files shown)


mvdan:
Sorry, what is this fixing? It would be best if you opened a bug report first, following the template to include a way to reproduce the issue.

bozaro:
@mvdan, this PR fixes very old #849 issue.
I also add `cue/testdata/eval/issue849.txtar` testcase for this issue.

mvdan:
Thanks. Could you please expand your commit message to explain the fix? At least what you're doing isn't obvious to me :)

bozaro:
Changes:
 - move `Complete` after all files are parsed to ast (fix infinity loop at load stage): `cue/load/import.go` -> `cue/load/instances.go`;
 - build transitive dependencies after files registered (fix infinity recursion at build stage): `internal/core/runtime/build.go`;
 - add test for import cycle at load stage;
 - add test for import cycle at build stage.
 
If revert changes from `cue/load/import.go`, `cue/load/instances.go`, `internal/core/runtime/build.go` then both added tests are hangs.

mvdan:
Thank you! I'll take a look. Out of curiosity, does this happen to fix https://github.com/cue-lang/cue/issues/2284 as well, or do you think it's a different issue altogether?

bozaro:
Looks like this PR also fixes #2284 issue, although I don't see a loop in the test code.

mvdan:
There is no loop in that issue - just one package that imports another package with the same name. Great to hear it's also fixed! I assume you also checked that this last added test also fails without your fix?

bozaro:
> Great to hear it's also fixed! I assume you also checked that this last added test also fails without your fix?

Yes. Without this PR test stuck on load stage.

mvdan:
Thanks! The changes look fine to me; do you want to squash your commits and write a good commit message explaining your changes and including all the "fixes" lines? In general we follow the guidelines at https://github.com/golang/go/wiki/CommitMessage.

bozaro:
Squashed :)

mvdan:
Thank you! I'll import into Gerrit and merge once Marcel has skimmed the changes, since I'm not super familiar with the package.

rogpeppe:
Although this does technically fix the issue, it also makes cyclic imports work, which raises another question: _should_ a cyclic import produce an error or not?

Currently the spec is [clear on the matter](https://cuelang.org/docs/references/spec/#import-declarations):
> It is illegal for a package to import itself, directly or indirectly

So this PR isn't quite right, according to that.

It's possible we might want to change the spec, but that's a larger policy question.
For now, it would be better to make this an explicit error, I think.

bozaro:
I couldn't figure out how to do a check for cyclic import without breaking the https://github.com/cue-lang/cue/issues/2284 fix.
From my point of view, allowing cyclic imports is less evil than an infinite loop at the loading stage.

bozaro:
Looks like #2284 is differ bug. But this PR mask it.

bozaro:
@rogpeppe, I created new PR with add packge import cycle error: https://github.com/cue-lang/cue/pull/2469

mvdan:
Roger is right - I had not noticed that this was allowing cyclic package imports to work, which shouldn't per the documentation. Thanks for opening the other PR! We can focus on that one, then.

PR #2469, opened 2023-07-03
cue/load: add package import cycle error

### LINKED ISSUE #849: cue/load: package cycles are not detected

bozaro:
Fixes #849
Closes #2435

### DIFF
--- a/cue/load/import.go
+++ b/cue/load/import.go
@@ -51,6 +51,19 @@ import (
 //	_       anonymous files (which may be marked with _)
 //	*       all packages
 func (l *loader) importPkg(pos token.Pos, p *build.Instance) []*build.Instance {
+	retErr := func(errs errors.Error) []*build.Instance {
+		// XXX: move this loop to ReportError
+		for _, err := range errors.Errors(errs) {
+			p.ReportError(err)
+		}
+		return []*build.Instance{p}
+	}
+
+	for _, item := range l.stk {
+		if item == p.ImportPath {
+			return retErr(&PackageError{Message: errors.NewMessagef("package import cycle not allowed")})
+		}
+	}
 	l.stk.Push(p.ImportPath)
 	defer l.stk.Pop()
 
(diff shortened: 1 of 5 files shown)


cue/load/import.go
```diff
@@ -51,6 +51,19 @@ import (
...
+		// XXX: move this loop to ReportError
+		for _, err := range errors.Errors(errs) {
+			p.ReportError(err)
+		}
+		return []*build.Instance{p}
+	}
+
+	for _, item := range l.stk {
+		if item == p.ImportPath {
+			return retErr(&PackageError{Message: errors.NewMessagef("import cycle not allowed")})
```
mvdan: nit: I would say "package import cycle not allowed" to make sure users are not confused with other types of cycles, e.g. modules or values.

mvdan (approved):
Thanks! Just one nit.

cc @rogpeppe

PR #2528, opened 2023-08-07
encoding/openapi: add ability to add additional schema constraints

howardjohn:
This makes https://github.com/cue-lang/cue/discussions/2522 possible. 

Basically we add a Value -> Raw OpenAPI schema. This allows users to add arbitrary addition OpenAPI constraints.

The existing Cue -> OpenAPI validation provides a solid base, but not that many validations can be expressed. This gives full flexibility.

In the PR I added a test that shows an example deriving an additional validation from comments. This showcases a similar implementation to what is seen in [kubebuilder](https://book.kubebuilder.io/reference/markers/crd-validation.html) (which converts Go types to OpenAPI (as k8s CRD, specifically).

----

In our full use case, we have protobuf inputs and want OpenAPI outputs. CEL currently facilitates this through Protobuf -> CEL -> OpenAPI. Currently, our OpenAPI schema is very minimal, as this conversion doesn't have a lot of constraints. We want to add a bunch of constraints; currently we do this through a Kubernetes mutating webhook, but this is operationally complex, so doing it through OpenAPI directly is a big improvement.

The validations we will need to add are quite broad, likely we will be utilizing most of the spec. In addition, we will utilize Kubernetes specific OpenAPI extensions. Most notably, CEL (https://kubernetes.io/docs/reference/using-api/cel/).

Our end goal is to provide something like https://book.kubebuilder.io/reference/markers/crd-validation.html. This is some existing tooling for translating Go types -> OpenAPI. Additional validation constraints are added via comments. We intend to do something similar, just with the source as protobuf types.

---

One consideration is if we derive constraints from protobuf -> openAPI (as done here), or have the constraints encoded as Cue constraints in the intermediary (protobuf constrains -> cue constraints, cue constraints -> openAPI constraints).

The former approach was chosen in this PR for a few reasons:
* Less work, we only need one set of constraints
* Complexity is pushed to users of the library instead of in the core library. The core just provides an extension point. Note that while this is both a Pro and a Con, as users now have to do more work, the [official docs](https://cuelang.org/docs/integrations/openapi/#generate-openapi) already point to the Istio tooling where this will live, so if anyone else is using Cue for CRDs, they likely can pick this up in the same way
* The mapping of Cue<->OpenAPI constraints is not 1:1, so we will not be able to express everything we want). This would likely block our primary use cases which are all custom to Kubernetes.

Some more discussion in slack: https://cuelang.slack.com/archives/CMY132JKY/p1693581326826669

### DIFF
--- a/encoding/openapi/openapi.go
+++ b/encoding/openapi/openapi.go
@@ -15,7 +15,6 @@
 package openapi
 
 import (
-	internaljson "cuelang.org/go/internal/encoding/json"
 	"fmt"
 	"strings"
 
@@ -24,6 +23,7 @@ import (
 	"cuelang.org/go/cue/errors"
 	"cuelang.org/go/cue/token"
 	cuejson "cuelang.org/go/encoding/json"
+	internaljson "cuelang.org/go/internal/encoding/json"
 )
 
 // A Config defines options for converting CUE to and from OpenAPI.
@@ -63,7 +63,7 @@ type Config struct {
 	DescriptionFunc func(v cue.Value) string
 
 	// AdditionalSchemasFunc allows adding additional schema constraints for a certain field.
-	AdditionalSchemasFunc func(v cue.Value) []SchemaConstraint
+	AdditionalSchemasFunc func(v cue.Value) map[string]ast.Expr
 
 	// SelfContained causes all non-expanded external references to be included
 	// in this document.
@@ -84,13 +84,6 @@ type Config struct {
 	// OpenAPI Schema. It is an error for an CUE value to refer to itself
 	// if this option is used.
 	ExpandReferences bool
-}
-
-// SchemaConstraint defines a single entry into a schema.
-// An example would be `SchemaConstraint{Key: "minItems", Attribute: 1}`
-type SchemaConstraint struct {
-	Key       string
-	Attribute ast.Expr
 }
 
 type Generator = Config
@@ -147,7 +140,6 @@ func toCUE(name string, x interface{}) (v ast.Expr, err error) {
 			"openapi: could not encode %s", name)
 	}
 	return v, nil
-
 }
 
 func (c *Config) compose(inst cue.InstanceOrValue, schemas *ast.StructLit) (x *ast.StructLit, err error) {
--- a/encoding/openapi/openapi.go
+++ b/encoding/openapi/openapi.go
@@ -15,6 +15,7 @@
 package openapi
 
 import (
+	internaljson "cuelang.org/go/internal/encoding/json"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


howardjohn:
Is there some process I need to go through to get a review on this?

encoding/openapi/openapi.go
```diff
@@ -62,6 +62,9 @@ type Config struct {
 	// the empty string is returned.
 	DescriptionFunc func(v cue.Value) string
 
+	// AdditionalSchemasFunc allows adding additional schema constraints for a certain field.
```
rogpeppe, line 65: Rather than saying what it allows, I'd suggest saying how it works.
For example (using guessed-at semantics):

```
// AdditionalSchemasFunc is called for every definition in the spec when encoding CUE
// to OpenAPI. If it returns a non-empty slice, the constraints specified in the slice will
// be associated with that definition in the generated OpenAPI schema.
```

Also, out of interest, why not return a map (e.g. `map[string] ast.Expr`) rather than a slice. Does ordering matter?

encoding/openapi/openapi.go
```diff
@@ -83,6 +86,13 @@ type Config struct {
 	ExpandReferences bool
 }
 
+// SchemaConstraint defines a single entry into a schema.
+// An example would be `SchemaConstraint{Key: "minItems", Attribute: 1}`
```
rogpeppe: This example isn't valid because the constant `1` isn't an `ast.Expr`.

Also, using an `ast.Expr` allows the encoding of arbitrary CUE expressions: it would be good to describe how they are encoded into the resulting OpenAPI schema.

encoding/openapi/openapi.go
howardjohn: I think map is probably a good call, I will update this

encoding/openapi/openapi.go
howardjohn: I really just want something that can be be converted into a json expression. Do you think its better to use a `map[?]?` or similar (not sure what type exactly), or use `ast.Expr` but document it better?

howardjohn: Done

encoding/openapi/openapi.go
howardjohn: Done

howardjohn:
One major hurdle I am running into is something like:

```
#HTTPRouteDestination: {
	headers:     [...#Headers]
}
// validateme:"message"
#Headers: {
    name: string
}
```

We end up building something that looks like
```
[...{
	name: string
}] [] path=[#HTTPRouteDestination headers]
```

So we aren't actually building `#Headers` really, but more the anonymous struct we happened to assign to `#Headers` (terminology may be off).

Now, we do also later build `#Headers`, but with ExpandReferences=true it doesn't matter, that ends up being discarded.

howardjohn:
We decided to move off Cue for this use case (https://github.com/istio/tools/pull/2674). Nothing against Cue, but using it strictly as intermediate state was causing more headaches than it was worth; direct translation from protobuf -> OpenAPI gives us a lot simpler codepath with more flexibility.

I think this PR is pretty hyper-focused only for our use case, so probably makes sense to close this. If anyone wants to pick this up though, feel free to. Thanks for all the help

PR #2532, opened 2023-08-09
encoding/gocode: Generate no longer requires deprecated *cue.Instance

alecholmez:
This PR attempts to tackle the use of deprecated APIs in the `encoding/gocode` Generate lib.
The current implementation relies on old `*cue.Runtime/*cue.Instance` declaration that have since been
superseded with `*build.Instance` and `*cue.Value`. The function signature of generate has been changed to use
a backwards compatible `cue.InstanceOrValue` interface.


alecholmez:
Here is an example of what is looks like from a users perspective:
```go
func main() {
	// Array of target directories which eval into CUE (including the CWD)
	// Can be globals, an app stack, core configs, or policies
	instances := load.Instances([]string{"test.cue"}, &load.Config{
		Dir: ".",
	})

	for i, instance := range instances {
		// check for errors on the instance
		// these are typically parsing errors
		if instance.Err != nil {
			log.Fatal(instance.Err.Error())
			continue
		}

		b, err := gocode.Generate(".", cuecontext.New().BuildInstance(instance), &gocode.Config{})
		if err != nil {
			log.Fatal(err)
		}

		err = os.WriteFile("cue_gen"+fmt.Sprint(i)+".go", b, 0644)
		if err != nil {
			log.Fatal(err)
		}
	}
}
```

cue/marshal.go
```diff
@@ -128,19 +128,20 @@ func (r *Runtime) Unmarshal(b []byte) ([]*Instance, error) {
...
 	staged := []instanceData{}
 	done := map[string]int{}
 
 	var errs errors.Error
 
-	var stageInstance func(i *Instance) (pos int)
-	stageInstance = func(i *Instance) (pos int) {
-		if p, ok := done[i.ImportPath]; ok {
+	var stageInstance func(i InstanceOrValue) (pos int)
+	stageInstance = func(i InstanceOrValue) (pos int) {
```
mvdan, line 138: this is a local func, so it should take `Value` directly

cue/marshal.go
```diff
@@ -194,14 +196,16 @@ func (r *Runtime) Marshal(instances ...*Instance) (b []byte, err error) {
 			if i == nil || !strings.Contains(imp, ".") {
 				continue // a builtin package.
 			}
-			stageInstance(i)
+
+			val := i.Value()
+			stageInstance(&val)
```
mvdan, line 201: you shouldn't need a pointer to a value, I think

cue/marshal.go
```diff
@@ -128,19 +128,20 @@ func (r *Runtime) Unmarshal(b []byte) ([]*Instance, error) {
...
 
 	var errs errors.Error
 
-	var stageInstance func(i *Instance) (pos int)
-	stageInstance = func(i *Instance) (pos int) {
-		if p, ok := done[i.ImportPath]; ok {
+	var stageInstance func(i InstanceOrValue) (pos int)
+	stageInstance = func(i InstanceOrValue) (pos int) {
+
+		if p, ok := done[i.Value().BuildInstance().ImportPath]; ok {
```
mvdan, line 140: I would call `.BuildInstance()` once and reuse the value throughout the func body, since it's not a free method call

mvdan (approved):

mvdan:
Apologies for the slowness on my part - looks good to me with some minor input :) Please rebase as well, since there appears to be one conflict.

mvdan:
Imported a modified version of this PR into Gerrit, per my comments above but also a few more tweaks, at https://review.gerrithub.io/c/cue-lang/cue/+/1174069.

PR #2554, opened 2023-08-18
gomod: update version

sivchari:
I've updated Go module version.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module cuelang.org/go
 
-go 1.20
+go 1.21
 
 require (
 	github.com/cockroachdb/apd/v3 v3.2.0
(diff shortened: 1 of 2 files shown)


mvdan:
Thanks for this PR! We want to support Go 1.20 until Go 1.22 is out, and the `go` directive in `go.mod` signals the language version we are using, so it could easily lead to us breaking people on Go 1.20 when we still want to support them in v0.7. I believe the Go toolchain also takes this version as a signal for a minimum version to use, so that wouldn't be a good signal to have either.

See https://github.com/cue-lang/cue/discussions/1937 as well.

sivchari:
@mvdan 
Thanks for quickly comment ! I see and I appreciate that you said. Btw, do you plan v0.8 to support Go.1.22.0 ?

mvdan:
CUE master should generally support Go master, since multiple of us use Go master as a daily driver. Whenever we release a CUE stable version, it should also support the latest Go stable version.

sivchari:
Nice and thanks for good work. CUE helps my works 🥇

Issue #2566, opened 2023-08-24
evaluator: it looks like the result of executing functions is not reused


bozaro:
### What version of CUE are you using (`cue version`)?

<pre>
$ cue version

cue version v0.0.0-20230816083946-7d801ba406ea

go version go1.20.7
      -buildmode exe
       -compiler gc
     CGO_ENABLED 1
          GOARCH amd64
            GOOS linux
         GOAMD64 v1
             vcs git
    vcs.revision 7d801ba406ea82923c19e66ce6aa419fdb62b2e8
        vcs.time 2023-08-16T08:39:46Z
    vcs.modified true
</pre>

### Does this issue reproduce with the latest stable release?

Yes

### What did you do?

I evaluate cue file with breakpoint on line: https://github.com/cue-lang/cue/blob/v0.6.0/pkg/strings/pkg.go#L300

```cue
import "strings"

let foo = strings.HasPrefix("foo", "bar")
a: foo
b: foo
c: foo
d: foo
e: foo
```

### What did you expect to see?

I expect only one interrupt at breakpoint (function `strings.HasPrefix` executed one time).

### What did you see instead?

I got 6 interrupts at breakpoints (function `strings.HasPrefix` executed 6 times).

mvdan:
Perhaps related to https://github.com/cue-lang/cue/issues/2243, where "chained" calls to builtins would result in increasingly slower evaluation.

bozaro:
I checked behaviour before and after changeset https://review.gerrithub.io/c/cue-lang/cue/+/549247 from #2243 and got exactly same behaviour.

bozaro:
Interestingly, in this example, `strings.ToLower` is executed exactly once:
```cue
import "strings"

let bar = strings.ToLower("BAR")
a: strings.HasPrefix("foo1", bar)
b: strings.HasPrefix("foo2", bar)
c: strings.HasPrefix("foo3", bar)
d: strings.HasPrefix("foo4", bar)
e: strings.HasPrefix("foo5", bar)
```
But in this example `strings.ToLower` is executed six times:
```cue
import "strings"

let bar = strings.ToLower("BAR")
a: bar
b: bar
c: bar
d: bar
e: bar
```

bozaro:
I spent several days debugging this issue and wrote the ugly workaround:
- https://github.com/cue-lang/cue/pull/2589

mvdan:
[...]

Issue #2007, opened 2022-10-14
language: function signatures


myitcv:
This is an umbrella placeholder issue for all things related to function signatures in CUE.

For example, when we publish a proposal for function signatures, we will add a comment to this issue and link to the proposal discussion.

Therefore, please subscribe to this issue if you are interested in following progress on this topic.

cc @rogpeppe @4ad

verdverm:
Not necessarily web assembly, but related, it would be cool if there was a way to use the function signatures wrapped around running a container to do the work. Probably only limited to the scripting layer.

Maybe this will be possible without any extra effort once the function syntax is available?

mvdan:
For those following this thread, please see our proposal and experimental support for CUE functions: https://github.com/cue-lang/cue/discussions/4484

Please leave any feedback on that discussion thread. We will keep this umbrella issue open for the time being to track the work.

PR #2587, opened 2023-09-11
cue/stats: add metric for builtin calls

bozaro:
Add `Builtins` metric to allow tests for issues like https://github.com/cue-lang/cue/issues/2566

### DIFF
--- a/internal/core/adt/expr.go
+++ b/internal/core/adt/expr.go
@@ -1481,6 +1481,7 @@ func (x *CallExpr) evaluate(c *OpContext, state vertexStatus) Value {
 	if b.IsValidator(len(args)) {
 		return &BuiltinValidator{x, b, args}
 	}
+	c.stats.Builtins++
 	result := b.call(c, pos(x), false, args)
 	if result == nil {
 		return nil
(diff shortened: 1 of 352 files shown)


internal/core/adt/expr.go
```diff
@@ -1481,6 +1481,7 @@ func (x *CallExpr) evaluate(c *OpContext, state vertexStatus) Value {
 	if b.IsValidator(len(args)) {
 		return &BuiltinValidator{x, b, args}
 	}
+	c.stats.Builtins++
```
mvdan, line 1484: note that BuiltinValidator ends up calling the `call` method as well; you should probably increment the counter inside that method to capture all builtin calls being performed.

mvdan (commented):
I'll run this past Marcel to see if he agrees that having builtin call counters in the stats is useful. I think it can be helpful when debugging or testing, like in your test in https://github.com/cue-lang/cue/pull/2588, so I don't oppose adding this stat.

We might want this metric counter in a slightly different form, like `Calls`. Once we add function signatures to the language per https://github.com/cue-lang/cue/issues/2007, perhaps implemented by wasm, then those could be counted the same way. I don't think splitting builtins calls from other function calls would be necessary.

Issue #2596, opened 2023-09-15
Can't build *adt.Vertex from same *build.Instance in multiple goroutines


bozaro:
### What version of CUE are you using (`cue version`)?

<pre>
$ cue version

cue version v0.0.0-20230912145313-0e9584387f3f

go version go1.21.1
      -buildmode exe
       -compiler gc
  DefaultGODEBUG panicnil=1
     CGO_ENABLED 1
          GOARCH amd64
            GOOS linux
         GOAMD64 v1
             vcs git
    vcs.revision 0e9584387f3fa0fa6b0a5c509d75c0088481df96
        vcs.time 2023-09-12T14:53:13Z
    vcs.modified true
</pre>

### Does this issue reproduce with the latest stable release?

Yes.

### What did you do?

I execute code like from multiple goroutines:
```go
		func(instance *build.Instance) {
			_, err := runtime.New().Build(nil, instance)
			if err != nil {
				t.Fatal(err)
			}
		}
```

Runnable testcase:
 - https://github.com/cue-lang/cue/pull/2597

### What did you expect to see?

No panics.

### What did you see instead?

```
=== RUN   TestX
panic: runtime error: index out of range [2] with length 2

goroutine 131 [running]:
cuelang.org/go/internal/core/runtime.resolveFile(0xc000782070, 0xc000133f80, 0xc0002c4b40, 0x4?)
	/home/bozaro/joom/cue/internal/core/runtime/resolve.go:157 +0xf85
cuelang.org/go/internal/core/runtime.(*Runtime).ResolveFiles(0xd469d8?, 0xc0002c4b40)
	/home/bozaro/joom/cue/internal/core/runtime/resolve.go:52 +0x1b7
cuelang.org/go/internal/core/runtime.(*Runtime).Build(0xc000780018, 0x0, 0xc0002c4b40)
	/home/bozaro/joom/cue/internal/core/runtime/build.go:69 +0x19c
cuelang.org/go/internal/core/runtime.(*Runtime).buildSpec(0xc000780018, 0x41ae6b?, 0xc0002c4960, 0xc00040c870)
	/home/bozaro/joom/cue/internal/core/runtime/build.go:150 +0x2cc
cuelang.org/go/internal/core/runtime.(*Runtime).Build.func1(0x0?)
	/home/bozaro/joom/cue/internal/core/runtime/build.go:64 +0x8a
[...]

PR #2597, opened 2023-09-15
Reproducing https://github.com/cue-lang/cue/issues/2596 issue

bozaro:
Modification for reproducing https://github.com/cue-lang/cue/issues/2596 issue

### DIFF
--- a/internal/core/runtime/resolve.go
+++ b/internal/core/runtime/resolve.go
@@ -15,8 +15,10 @@
 package runtime
 
 import (
+	"math/rand"
 	"path"
 	"strconv"
+	"time"
 
 	"cuelang.org/go/cue/ast"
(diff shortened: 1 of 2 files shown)


internal/core/runtime/resolve.go
```diff
@@ -152,6 +154,7 @@ func resolveFile(
 			u.Scope = f
 			continue
 		}
+		time.Sleep(time.Millisecond * time.Duration(rand.Intn(10)+1))
```
bozaro, line 157: A small delay to increase the reproducibility of the race.

mvdan:
Thanks for the PR to help reproduce the race; I think leaving this open as a draft is fine for now. We can revisit when we look at making the API safe for concurrent use.

Issue #943, opened 2021-07-03
Proposal: core builtin extensions


cueckoo:
_Originally opened by @mpvl in https://github.com/cuelang/cue/issues/943_

## Definitions
Before we introduce some of the proposed builtins, we formally introduce some as-of-yet undocumented language features.

### Functions

We propose cue supports named argument functions and calls to “structs” as a shorthand for the common macro pattern (e.g.  `(s & { _, a: x}).out`).

A function argument is now defined as:
```
	Argument       = [ identifier ":" ] Expression .
```
Any named argument must be followed by other named arguments.

The expression `s(a: x, b: y)`, where `s` is a struct, is now a shorthand for `s & {_, a: x, b: y)`.

### Validator

A validator is a special builtin that is evaluated by unifying it with other values whereby the result is one of a few outcomes:

* pass: returns `_` if the validation is successful and making the value with which it was unified more specific does not change this result (or it is a final evaluation).
* incomplete error: the validation failed, but making the value with which it was unified more specific could change this result
* fatal error: the validation failed and making the value more specific cannot change this result.

A validator _must_ be run at the last stage of evaluating a node, after a fixed point is reached evaluating all all non-validator values, in which case any error is considered a fatal error. A validator _may_ be run at earlier stages of the evaluation of a node, in which case an incomplete error signifies that the decision on validity must be postponed.

An example of a language-level validator is `<10`. `struct.MinFields` and `struct.MaxFields` are examples of validators of builtin packages.

Validators can be thought of as a Go function that has an error return signature.

### Inferred validators

Optional: Builtin functions that have the signature `foo(x1, x2, …, xn) bool` may be implicitly interpreted as validators of the signature `foo(x2, …, xn) error`.

### The CUE function notation
We define the following signature format for cue functions:
```
FunctionDecl   = identifier Arguments "::" Expression .   
Arguments      = "(" [ Argument { "," Argument } [ "," ] ] ")" .
Argument       = [ identifier ":" ] Expression .
```
Either all or none of the arguments should be named.

The following rules apply for calling functions with this signature:

* An argument with a default value in its expression may be omitted in a call. All other arguments must be present in a call.
* A call must either have all named, or all unnamed arguments. This could be rel

These rules could be relaxed later.
### builtins to replace `_|_` (bottom)

Although `_|_` is part of the standard CUE idiom, it has several issues:

* no ability to associate user-defined message to bottom
* meaning of comparison against bottom is unclear
* the symbol looks offensive to some

[...]

PR #2605, opened 2023-09-20
internal/core/adt: add `isconcrete` builtin func

bozaro:
This PR implements `isconcrete` buildin func from https://github.com/cue-lang/cue/issues/943 proposal.

isconcrete(expr) :: bool

`isconcrete` reports whether expr resolves to a concrete value, returning true if it does and false otherwise. It is a fatal error if an expression can never evaluate to true.

**Motivation**: replaces `if a.foo != _|_ {` field value check without hiding the underlying error.

### DIFF
--- a/cue/testdata/builtins/isconcrete.txtar
+++ b/cue/testdata/builtins/isconcrete.txtar
@@ -1,0 +1,106 @@
+-- in.cue --
+a: {}
+b: int
+
+c: isconcrete(a)   // true
+d: isconcrete(b)   // false
+e: isconcrete(a.b) // false(b could still be defined)
+f: isconcrete(b.c) // fatal error (b.c can never be satisfied)
+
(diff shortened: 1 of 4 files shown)


fab29p:
With the following file, there is a difference in the behavior of isconcrete and the behavior of `!= _|_` when doing a `cue export`. Try by commenting the line with `a_1` then the line with `a_2`. The test on `a_1` raises an error. 

~~~
a: "z"
x: {
    y: 1
    
}

let u = {
    for k, v in x {
        (k): v
    }
}

a_1: isconcrete(u[a])
a_2: u[a] != _|_
~~~

fab29p:
Notice that 

* the proposal for `exists` exhibit the same behavior
* if `let u=` is replaced by `u:` then the problem disappear

PR #2622, opened 2023-09-28
protobuf: enable protobuf standard `required` mechanism

howardjohn:
See https://google.aip.dev/203. Currently, there is support for an option `(cue.opt).required`. This is fine, but requires importing the `cue.proto`, and exposes cue to the public facing API.

With this approach, the built-in field behavior mechanism can be used, which is the standard way to declare a required field in proto.

### DIFF
--- a/encoding/protobuf/parse.go
+++ b/encoding/protobuf/parse.go
@@ -796,7 +796,10 @@ func (p *optionParser) parse(options []*proto.Option) {
 			if !p.required {
 				constraint.Optional = token.NoSpace.Pos()
 			}
-
+		case "(google.api.field_behavior)":
+			if o.Constant.Source == "REQUIRED" {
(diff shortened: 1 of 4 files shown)


mvdan:
Apologies that we didn't respond here sooner. Is there a particular reason you closed this PR?

Personally this seems like a good change to incorporate, so if you want to re-open the PR I can import it to Gerrit and get it merged. I can always do that even if you don't reopen the PR, but ideally I'd like your permission to do so :)

howardjohn:
Feel free to pick it up. We ended up moving off of Cue so there wasn't a
need to pursue this further

mvdan:
Thanks! Imported to Gerrit as https://review.gerrithub.io/c/cue-lang/cue/+/1172314.

PR #2643, opened 2023-10-10
chore: gofumpt

uhthomas:
Format and simplify code inline with gofumpt:

```sh
gofumpt -extra -w .
```

### DIFF
--- a/cmd/cue/cmd/add.go
+++ b/cmd/cue/cmd/add.go
@@ -203,7 +203,7 @@ func (fo *originalFile) restore() error {
 	if len(fo.contents) == 0 {
 		return os.Remove(fo.filename)
 	}
-	return os.WriteFile(fo.filename, fo.contents, 0644)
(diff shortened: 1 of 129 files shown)


mvdan:
I'm obviously biased in favor of a change like this :) We need to be careful with big reformats or automated edits like these, though. Right now is a tricky time since Marcel has a large in-progress stack of changes for the performance work in v0.7, so a big patch like this is extremely likely to cause him dozens of merge conflicts and slow him down.

For the time being, we do enforce gofmt. I imagine that enforcing gofumpt too wouldn't be controversial with the other maintainers - I will bring it up once master is in less of a tricky spot, probably in a few weeks. Hopefully that makes sense :)

mvdan:
Given the increasing amount of conflicts, and per my comment above, I'm going to close this for now. I'm happy to talk about stricter formatting once the big refactors have landed, though - perhaps we should use an issue or discussion before we jump straight to a PR, given the high chance for conflicts. Thanks!

Issue #826, opened 2021-07-03
cmd/cue: multi-byte characters in YAML result in bad formatting


cueckoo:
_Originally opened by @uhthomas in https://github.com/cuelang/cue/issues/826_

### What version of CUE are you using (`cue version`)?

<pre>
$ cue version
cue version v0.3.0-beta.5 linux/amd64
</pre>

### Does this issue reproduce with the latest release?

Yes.

### What did you do?

```sh
curl -Lfv https://github.com/jetstack/cert-manager/releases/download/v1.2.0/cert-manager.yaml > t.yaml
cue import --list t.yaml
```

### What did you expect to see?

A nicely formatted CUE file.

### What did you see instead?

![](https://kipp.6f.io/n4JI0s8Cg8rH.png)

cueckoo:
_Original reply by @uhthomas in https://github.com/cuelang/cue/issues/826#issuecomment-797121973_

Another example

![](https://kipp.6f.io/g7nBt0f3f_Pd.png)

cueckoo:
_Original reply by @myitcv in https://github.com/cuelang/cue/issues/826#issuecomment-797879468_

Confirmed with 7c5d28e6721a260dc2efdf623a015b5aea195d9e

mpvl:
Formatting looks fine to me. A more specific example is welcome.

uhthomas:
@mpvl Sorry, I must not have seen your comment.

Given https://github.com/cert-manager/cert-manager/releases/download/v1.11.0/cert-manager.crds.yaml

```sh
❯ cue import --list cert-manager.yaml
```

Produces https://gist.github.com/uhthomas/da3842ef5880be0e26f05a00272a3ae4

The formatting is initially fine, but then looks really weird around here: https://gist.github.com/uhthomas/da3842ef5880be0e26f05a00272a3ae4#file-cert-manager-cue-L3081

```cue
apiVersion:
	"apiextensions.k8s.io/v1", kind:
	"CustomResourceDefinition", metadata: {

	name: "certificaterequests.cert-manager.io"
	labels: {
		app:                          "cert-manager"
		"app.kubernetes.io/name":     "cert-manager"
		"app.kubernetes.io/instance": "cert-manager"
		// Generated labels
		"app.kubernetes.io/version": "v1.11.0"
	}
}, spec: {

	group: "cert-manager.io"
```

I hope it's clear that something is wrong here.

PR #2645, opened 2023-10-11
fix(format): indent list elements correctly

### LINKED ISSUE #2314: cue/fmt: Lists are indented too much

<pre>
$ cue version
0.4.3-1
</pre>

Yes.

### What did you do?

```cue
import "list"

#List: items: list.Concat([
    a,

    b
])
```

```cue
import "list"

#List: items: list.Concat([
    a,

    b
])
```

### What did you see instead?

```cue
import "list"

#List: items: list.Concat([
        a,

    b
])
```

uhthomas:
Vertical tabs are not inserted when the next expression needs a form feed, but are erroneously inserted for expressions which hide the underlying type. This is fixed by recursively checking if expressions need a form feed.

Fixes: #2314

### DIFF
--- a/cmd/cue/cmd/testdata/script/get_go_types.txtar
+++ b/cmd/cue/cmd/testdata/script/get_go_types.txtar
@@ -266,12 +266,12 @@ import (
 	Slice1: [...int] @go(,[]int)
 	Slice2: [...] @go(,[]interface{})
 	Slice3?: null | [...] @go(,*[]json.Unmarshaler)
-	Array1:  5 * [int]    @go(,[5]int)
-	Array2:  5 * [_]      @go(,[5]interface{})
+	Array1: 5 * [int] @go(,[5]int)
+	Array2: 5 * [_] @go(,[5]interface{})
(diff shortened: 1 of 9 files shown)


uhthomas:
Interestingly, this also fixes #826.

mvdan (approved):
This does mean that we vertically align fields in fewer cases, but some of those were clearly broken like you show, so this definitely seems like a good step.

We can always tweak the heuristic to vertically align more often again in the future, bearing in mind to not break your added test cases.

mvdan:
Imported this PR into Gerrit as https://review.gerrithub.io/c/cue-lang/cue/+/1170966, since GitHub is a git mirror.

I can't review commit messages on GitHub PRs like on Gerrit, so I took the liberty to expand your commit message a bit to explain what I think is going on with the problem and your fix. Hope that's okay. You can always leave a review on the Gerrit CL if you would like it written differently :)

mvdan:
Also FYI, CI failed since `go generate ./...` resulted in a slight change in generated CUE. Fixed that in the CL.

Issue #2691, opened 2023-11-11
cue/cmd: `import crd` for importing Kubernetes CustomResourceDefinitions


justenstall:
## Is your feature request related to a problem? Please describe.

CUE cannot import [Kubernetes CustomResourceDefinitions (CRDs)](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#customresourcedefinitions). Importing CRDs into CUE would be a huge value add for Kubernetes users.

## Describe the solution you'd like

I would like a `cue import crd` command that is able to import a YAML or JSON encoded CustomResourceDefinition to CUE.

### `cue import openapi`

Since CRDs use a variation of OpenAPIv3 to define their schema, `cue import openapi` can be used to import the schema with some extra help from a query tool like yq. Here is an example script to extract a valid OpenAPI definition from the CRD with yq and then import it as CUE:

```sh
CRD="<file>.yaml"

yq '{"openapi": "3.0.0"} *
	{"info": {"title": .spec.group}} *
	{"info": {"version": .spec.versions[0].name}} *
	{"components": { "schemas": { .spec.names.kind: .spec.versions[0].schema.openAPIV3Schema}}}' \
	"$CRD_FILE" > crd-openapi.yaml 

cue import openapi crd-openapi.yaml
```

This is not a viable solution because of Kubernetes' OpenAPIv3 extensions explained in the _Additional Context_ section below.

### `cue get go`

Many CRDs are created with a code generation tool such as [`controller-gen`](https://book.kubebuilder.io/reference/controller-gen) (from the [kubebuilder SDK](https://book.kubebuilder.io/)), which generates the CRD from annotated Go code. I have used `cue get go` to import the CRD from the Go types, but it is not a 100% match for the `controller-gen`-created schema. Adding direct support for CRD --> CUE is ideal.

> See also: #2679 

### OpenAPIv3 vs Structural Schema

Kubernetes calls its flavor of OpenAPIv3 "structural schema", which is defined as follows:

> A structural schema is an [OpenAPI v3.0 validation schema](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation) which:
>
> 1. specifies a non-empty type (via type in OpenAPI) for the root, for each specified field of an object node (via properties or additionalProperties in OpenAPI) and for each item in an array node (via items in OpenAPI), with the exception of:
>    - a node with x-kubernetes-int-or-string: true
[...]

PR #2701, opened 2023-11-21
cue/cmd: `import crd` for importing Kubernetes CustomResourceDefinitions

justenstall:
Updates #2691


stefanprodan:
Hey @justenstall I've found a serious issue with CRD import in Timoni, you may want to port the fix here: https://github.com/stefanprodan/timoni/pull/279

rogpeppe:
Thanks for this! However, we chose to go with an alternative approach that integrated CRD support more deeply into the `encoding/jsonschema` package, as implemented in https://review.gerrithub.io/c/cue-lang/cue/+/1217836, so closing this as redundant now. Please raise issues if that approach falls short of your needs/expectations.

PR #2702, opened 2023-11-22
cue/cmd: Prevent exec.Run fatal errors for non-zero exit codes

### LINKED ISSUE #2632: cmd/cue: exec.Run always gives fatal error on non-zero exit status

<pre>
$ cue version
go version devel go1.21-af8f94e3c5 Tue Jul 11 21:30:51 2023 +0000
      -buildmode exe
       -compiler gc
  DefaultGODEBUG panicnil=1
     CGO_ENABLED 1
          GOARCH amd64
            GOOS linux
         GOAMD64 v1
             vcs git
    vcs.revision 6c71b2c90d63846f2983deb07102253298a568c0
        vcs.time 2023-10-04T17:02:06Z
    vcs.modified false

</pre>

Yes

### What did you do?

```
exec cue cmd foo
stdout 'result'

-- cue.mod/module.cue --
module: "example.com"

-- foo_tool.cue --
package foo
import (
	"tool/exec"
	"tool/cli"
)

command: foo: {
	first: exec.Run & {
		cmd: ["false"]
	}
	second: cli.Print & {
		$after: first
		text: "result \(first.success)"
	}
}
```

A test pass. The `second` task should run even when `first` has failed.

From the [docs](https://pkg.go.dev/cuelang.org/go/pkg/tool/exec):

> success is set to true when the process terminates with a zero exit code or false otherwise. The user can explicitly specify the value [to] force a fatal error if the desired success code is not reached.

This seems to imply that if `success` is not explicitly set, the error is not fatal.

### What did you see instead?

```
> exec cue cmd foo
[stderr]
task failed: command "false" failed: exit status 1
[exit status 1]
FAIL: /tmp/testscript3963459195/x.txtar/script.txtar:1: unexpected command failure
```

The error from the non-zero exit status seems to be fatal, aborting execution of the command.

This makes it hard to make subsequent actions conditional on a command failing.

nickfiggins:
This updates the logic in exec.Run to not return an error if a command returns a non-zero exit status. As a result, the success field can be used to determine if a command succeeded or failed.

Any errors that aren't related to the command failing to run or not completing successfully are still returned.

Fixes #2632

### DIFF
--- a/pkg/tool/exec/exec.go
+++ b/pkg/tool/exec/exec.go
@@ -71,6 +71,16 @@ func (c *execCmd) Run(ctx *task.Context) (res interface{}, err error) {
 		cmd.Stderr = ctx.Stderr
 	}
 
+	// TODO(mvdan): exec.Run declares mustSucceed as a regular field with a default of true.
+	// We should be able to rely on that here, removing the need for Exists and repeating the default.
+	mustSucceed := true
+	if v := ctx.Obj.LookupPath(cue.ParsePath("mustSucceed")); v.Exists() {
+		mustSucceed, err = v.Bool()
+		if err != nil {
+			return nil, errors.Wrapf(err, v.Pos(), "invalid bool value")
+		}
+	}
+
 	update := map[string]interface{}{}
 	if captureOut {
 		var stdout []byte
@@ -80,15 +90,23 @@ func (c *execCmd) Run(ctx *task.Context) (res interface{}, err error) {
 		err = cmd.Run()
 	}
 	update["success"] = err == nil
-	if err != nil {
-		if exit := (*exec.ExitError)(nil); errors.As(err, &exit) && captureErr {
+
+	if err == nil {
+		return update, nil
+	}
+
+	if captureErr {
+		if exit := (*exec.ExitError)(nil); errors.As(err, &exit) {
 			update["stderr"] = string(exit.Stderr)
 		} else {
-			update = nil
+			update["stderr"] = err.Error()
 		}
-		err = fmt.Errorf("command %q failed: %v", doc, err)
 	}
-	return update, err
+	if !mustSucceed {
+		return update, nil
+	}
+
+	return nil, fmt.Errorf("command %q failed: %v", doc, err)
 }
 
 func mkCommand(ctx *task.Context) (c *exec.Cmd, doc string, err error) {
--- a/cmd/cue/cmd/testdata/script/cmd_execerr.txtar
+++ b/cmd/cue/cmd/testdata/script/cmd_execerr.txtar
@@ -1,0 +1,56 @@
+! exec cue cmd cannotFail
+cmp stderr cannotFail-stderr.golden
+
+exec cue cmd canFail
+cmp stdout canFail-stdout.golden
+
+-- cue.mod/module.cue --
+module: "example.com"
+
+-- cannotFail-stderr.golden --
+task failed: command "false" failed: exit status 1
+-- canFail-stdout.golden --
+errExit:
+success=false
+hasStderr=false
+
+errNotFound:
+success=false
+hasStderr=true
+-- foo_tool.cue --
+package foo
+
+import (
+	"tool/cli"
+	"tool/exec"
+)
+
+command: cannotFail: {
+	one: exec.Run & {
+		cmd: ["false"]
+	}
+}
+
+command: canFail: {
+	errExit: exec.Run & {
+		cmd: ["false"]
+		stderr: string
+		mustSucceed: false
+	}
+	errNotFound: exec.Run & {
+		cmd: ["program-does-not-exist"]
+		stderr: string
+		mustSucceed: false
+	}
+	second: cli.Print & {
+		text: """
+			errExit:
+			success=\(errExit.success)
+			hasStderr=\(errExit.stderr != "")
+
+			errNotFound:
+			success=\(errNotFound.success)
+			hasStderr=\(errNotFound.stderr != "")
+			"""
+	}
+}
--- a/pkg/tool/exec/exec.cue
+++ b/pkg/tool/exec/exec.cue
@@ -48,4 +48,8 @@ Run: {
 	// code or false otherwise. The user can explicitly specify the value
 	// force a fatal error if the desired success code is not reached.
 	success: bool
+
+	// mustSucceed indicates whether a command must succeed, in which case success==false results in a fatal error.
+	// This option is enabled by default, but may be disabled to control what is done when a command execution fails.
+	mustSucceed: bool | *true
 }
--- a/pkg/tool/exec/pkg.go
+++ b/pkg/tool/exec/pkg.go
@@ -38,6 +38,10 @@
 //		// code or false otherwise. The user can explicitly specify the value
 //		// force a fatal error if the desired success code is not reached.
 //		success: bool
+//
+//		// mustSucceed indicates whether a command must succeed, in which case success==false results in a fatal error.
+//		// This option is enabled by default, but may be disabled to control what is done when a command execution fails.
+//		mustSucceed: bool | *true
 //	}
 package exec
 
@@ -62,10 +66,11 @@ var p = &pkg.Package{
 		env: {
 			[string]: string | [...=~"="]
 		}
-		stdout:  *null | string | bytes
-		stderr:  *null | string | bytes
-		stdin:   *null | string | bytes
-		success: bool
+		stdout:      *null | string | bytes
+		stderr:      *null | string | bytes
+		stdin:       *null | string | bytes
+		success:     bool
+		mustSucceed: bool | *true
 	}
 }`,
 }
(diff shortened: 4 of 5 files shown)


nickfiggins:
been learning cue and this issue sounded interesting, so wanted to give it a shot. Let me know if I misunderstood the intent here at all 😃

pkg/tool/exec/exec.go
```diff
@@ -80,15 +80,19 @@ func (c *execCmd) Run(ctx *task.Context) (res interface{}, err error) {
...
 	}
 	update["success"] = err == nil
-	if err != nil {
-		if exit := (*exec.ExitError)(nil); errors.As(err, &exit) && captureErr {
+
+	if err == nil {
+		return update, nil
+	}
+
+	if exit := (*exec.ExitError)(nil); errors.As(err, &exit) {
```
mvdan: I've never seen this single-line style for As before, interesting :) seems fine to me to be clear.

cmd/cue/cmd/testdata/script/cmd_errcode.txtar
```diff
@@ -16,4 +19,5 @@ command: errcode: {
 		kind:   "exec"
 		cmd:    "ls --badflags"
 		stderr: string // suppress error message
+		success: true
```
mvdan: Interesting, so any users relying on the current behavior where `cue cmd foo` fails if an `exec.Run` command returns a non-zero exit code would need to add a line like this.

That is technically a breaking change, and I wonder how many users it would affect in practice. Then again, the `success` field can't be of any use without a fix like this, like Roger pointed out in the original bug report, so I think we want to make the change anyway.

One thought outloud is that we could later also consider something like `exec.MustRun` which would be a shorthand for `exec.Run & {success: true}`.

mvdan (commented):
Thanks for this patch! This looks great in terms of implementation and testing :)

Let me run it by the others to confirm whether we are OK with the user-facing API change.

cmd/cue/cmd/testdata/script/cmd_errcode.txtar
rogpeppe: Another possibility might be to have another field that can be explicitly set to allow a failed command. It would default to the current behaviour.

For example, `allowFailure: *false | bool`, or maybe `mustSucceed: *true | bool`

Then if someone wants to allow a failed command they can do `allowFailure: true` or `mustSucceed: false` and the `success` field will indicate if the command did actually fail.

I think I might prefer this approach as not only does it preserve compatibility, but it also means that the default behaviour is to fail when a command fails, which seems like it might be the right default.

nickfiggins: Ah yeah - that makes sense to me in order avoid the breaking change and also keep the same default.  Happy to add that in if there's alignment here 🙂 

One thing I am thinking about is assuming we would still want to have a fatal error when `cmd.Run` returns an error that isn't of type `os.ExitError` (as done in this PR currently), I'm thinking `mustSucceed: false` might be clearer than `allowFailure: true` since the latter may make it seem like that command would never fail?  

However, if we want to disregard all errors when a flag is set then maybe setting `allowFailures: true` would make more sense. Let me know what you all think / if there's a preference around that.

jpluscplusm: Apologies if I'm missing something,  but may I suggest that the field not be a boolean, but an int instead - representing the command's exit code? 

That way, a default of "0" could be applied to keep the current backwards-compatible behaviour, whilst allowing the user control over which non-zero exit codes are permitted, via standard CUE number bounds, or a disjunction, or any other CUE-y mechanism.

mvdan: @nickfiggins please hold off on updating this PR - we had a brief internal discussion on what kind of API change we should make here, so I'll polish up those notes and share early next week :)

myitcv: > Apologies if I'm missing something, but may I suggest that the field not be a boolean, but an int instead - representing the command's exit code?

I believe the reason the `tool/exec` API does not expose the exit code is the same reason it doesn't appear in the https://pkg.go.dev/os/exec API: namely the behaviour around exit codes is OS-specific. @mvdan might well be able to offer more thoughts/colour on that point.

mvdan:
@nickfiggins apologies for the delay - we have discussed a number of options and to what degree we want to keep backwards compatibility here. `Run` is documented as follows, which would indicate that a command run isn't required to succeed:

> success is set to true when the process terminates with a zero exit code or false otherwise. The user can explicitly specify the value force a fatal error if the desired success code is not reached.

However, the currently implemented behavior has been in place for years, and some of those users may depend on it, so asking them to switch to a new API or set an option would be unfortunate and might catch some users by surprise. Moreover, going ahead, we think that the default _should_ be to require a command to succeed to continue the execution of `cue cmd`, much like CI systems like GitHub Actions run the Bash shell with `set -o errexit`, where any command failure stops the entire script.

So, following @rogpeppe's thoughts above, we're thinking of adding a boolean flag:

```
// Run continues to result in a fatal error if success==false, and we add the "mustSucceed" field.
Run: {
  // rest of fields unchanged, including "success"

  // mustSucceed indicates whether a command must succeed, in which case success==false results in a fatal error.
  // This option is enabled by default, but may be disabled to control what is done when a command execution fails.
  mustSucceed: bool | *true
}
```

Existing users would continue to do exactly the same, for example:

```
cannotFail: exec.Run & {cmd: "foo"}
```

And then, users who want to use the "success" boolean as either true or false can set "mustSucceed" to false:

```
canFail: exec.Run & {cmd: "bar", mustSucceed: false}
if canFail.success {
  // logic based on the resulting status
}
```

Let me know if the above is clear. If you are able to update the PR this week, we will attempt to include this change in the upcoming v0.7.0 release. Thanks!

cmd/cue/cmd/testdata/script/cmd_errcode.txtar
mvdan: Correct; exit codes are relatively common in Unix-like and POSIX-compliant systems, and Go does seem to support them on all platforms including Windows if you reach for the lower-level https://pkg.go.dev/os#ProcessState.ExitCode API, but they [aren't a portable standard](https://stackoverflow.com/questions/1538884/what-standard-application-return-exit-codes-should-an-application-support) that is well understood and should replace `success: bool`.

That said, we could still expose the exit code alongside the success boolean, if that is useful to some users who for example run on Linux. If anyone has need for that, please raise an issue and we can discuss a change.

nickfiggins:
@mvdan Thanks for all the detail and explanation, just pushed the changes 🙂 . Let me know if there's any issues

mvdan (approved):
Thank you! I think you can slightly simplify two bits of the code, but since we want to release v0.7.0-rc.1 today with this change, I hope it's okay with you if I make those changes as I import the PR into Gerrit for merging.

mvdan:
Looking more closely at your latest commit:

> Any errors that aren't related to the command failing to run or not completing successfully are always still returned.

I tend to disagree there; it still sounds useful to use `mustSucceed: false` to capture other types of exec errors, such as when a program isn't available like `cmd: "non-existent-program"`. It would also be rather surprising if `mustSucceed: false` allowed some failures to continue evaluation, but others not.

I'll tweak your code so that it treats non-ExitError errors with the same logic, and appends their error strings to `Run.stderr`, so that they may be showed to the end user or used in some way by the CUE logic. We could later add more fields to `Run` to differentiate "command failed to start" versus "command started and then exited with a non-zero exit code" errors, but for now that doesn't seem particularly important.

I'll add myself as co-author in the commit, for the sake of clarifying that there were non-trivial tweaks. I'll also push another commit here, so that one can diff between the two. I'm again hoping that this is OK with you, for the sake of including the fix in v0.7 :)

mvdan:
Done; you can look through https://github.com/cue-lang/cue/compare/5dd19902850e7d41fcd241c9d9ba539d992627f5..a454651883a739e0109c3399c7197e2908c11c58 to see my changes. There is a bit of other noise unfortunately, since I also rebased on master.

Importing to Gerrit now, for one last round of review.

nickfiggins:
> I tend to disagree there; it still sounds useful to use `mustSucceed: false` to capture other types of exec errors, such as when a program isn't available like `cmd: "non-existent-program"`. It would also be rather surprising if `mustSucceed: false` allowed some failures to continue evaluation, but others not.
> 

That totally makes sense, I was going to ask for clarity around that but figured there wasn't much time if it were to be included in the next release 🙂 .

> Done; you can look through https://github.com/cue-lang/cue/compare/5dd19902850e7d41fcd241c9d9ba539d992627f5..a454651883a739e0109c3399c7197e2908c11c58 to see my changes. There is a bit of other noise unfortunately, since I also rebased on master.
> 
> Importing to Gerrit now, for one last round of review.

Ah I see, this makes sense. Thank you!

mvdan:
I ended running out of time to cut a release today, so it will have to be monday. Perhaps there wasn't a need for me to rush merging this after all :) In any case, if you have any further thoughts about the design, or want to send a follow-up PR with any other changes, by all means please do!

Issue #2614, opened 2023-09-25
pkg/net: `IPv6` validator function


jpluscplusm:
**Is your feature request related to a problem? Please describe.**

I'd like to constrain fields to contain IPv6 addresses. I'm doing this in the context of structured data that represents DNS records, where records of type `AAAA` are only allowed to contain IPv6 addresses.

**Describe the solution you'd like**

I'd like a builtin validator(?) function to parallel https://pkg.go.dev/cuelang.org/go@v0.6.0/pkg/net#IPv4 (probably named `IPv6`), which operates similarly to `IPv4` by only permitting fields to which it's attached to be assigned IPv6 addresses. 

**Describe alternatives you've considered**

I could roll my own regex to validate IPv6 string forms.

**Additional context**

This field being missing seems like a mere oversight, seen in the context of the `net` package - which already has several IPv6-aware functions.

mvdan:
I think adding this API is reasonable, even if just for symmetry. Want to send a CL?

Worth noting that our `net` package is built atop Go's, so `IPv4` is using https://pkg.go.dev/net#IP.To4:

> If ip is not an IPv4 address, To4 returns nil.

However, upstream doesn't seem to have an easy way to check if an IP is IPv6; there is https://pkg.go.dev/net#IP.To16, but it seems to not require the IP to actually be IPv6. Look at the example with an IPv4, which works fine.

The answer might be to switch over to `net/netip` - it is meant to be faster, and it does seem to have a way to check if an address is IPv6: https://pkg.go.dev/net/netip#Addr.Is6

mvdan:
If I'm correct about the switch to `net/netip` being a good idea and perhaps even necessary to implement this, I would suggest to do it in multiple steps: first swap `net` for `net/netip`, and then add new CUE standard library APIs like `IPv6`.

PR #2706, opened 2023-11-23
pkg/net: replace net with net/netip

martingreber:
This replaces the go net package with net/netip as the latter is a better alternative and supports IPv6 checks.

Updates #2614

### DIFF
--- a/pkg/net/ip.go
+++ b/pkg/net/ip.go
@@ -17,7 +17,7 @@ package net
 
 import (
 	"fmt"
-	"net"
+	"net/netip"
 
 	"cuelang.org/go/cue"
 )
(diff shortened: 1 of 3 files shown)


mvdan (commented):
Thanks for the PR! Overall looks good.

I think it's fine to update the tests to expect the new error strings. We exposed the `net` errors as-is from Go, so if we switch to `net/netip`, we should continue to do that for the new package. If we wanted the errors to be particular to CUE, or stable over time, we would have to filter or replace all error values, but we don't really do or need to.

martingreber:
Cheers @mvdan. I've updated the tests and amended the commit to rely on the `net/netip` errors.

pkg/net/ip.go
```diff
@@ -16,10 +16,9 @@
 package net
 
 import (
-	"fmt"
-	"net"
-
 	"cuelang.org/go/cue"
+	"fmt"
+	"net/netip"
```
mvdan, line 20: please keep the std imports as a separate first group

pkg/net/ip.go
```diff
@@ -28,76 +27,87 @@ const (
...
 			return nil, err
 		}
-		_, gonet, err := net.ParseCIDR(s)
+		cidr, err := netip.ParsePrefix(s)
 		if err != nil {
 			return nil, err
 		}
-		return gonet, nil
+		if !cidr.IsValid() {
+			return nil, err
```
mvdan: this would result in `nil, nil`, right?

I suspect this shouldn't be possible; it would seem odd for ParsePrefix to return an invalid prefix with a nil error.

If it is possible, then I suspect what we want to do is return `nil, fmt.Errorf(...)`.

pkg/net/ip.go
```diff
@@ -28,76 +27,87 @@ const (
...
 			return nil, err
 		}
-		_, gonet, err := net.ParseCIDR(string(b))
+		cidr, err := netip.ParsePrefix(string(b))
 		if err != nil {
 			return nil, err
 		}
-		return gonet, nil
+		if !cidr.IsValid() {
+			return nil, err
```
mvdan: same as the other ParsePrefix comment.

mvdan (approved):
SGTM with some thoughts about the ParsePrefix error handling.

pkg/net/ip.go
martingreber: Good catch. `netip.ParsePrefix` returns an error rather than an invalid `Prefix` so that check is redundant. I'll remove it.

martingreber: Fixed.

pkg/net/ip.go
martingreber: Fixed.

martingreber:
BTW, I'll get a PR ready for the Cue `IPv6` check ready once this is merged. Figured that should be in a separate PR

mvdan:
Thank you! I'll be importing this PR to Gerrit now for merging, since GitHub is simply a git mirror for Gerrit. If you want to send further changes, I'd suggest sending them straight to Gerrit, for the sake of avoiding this extra step and speeding up the process :) https://github.com/cue-lang/cue/blob/master/CONTRIBUTING.md#preparing-for-gerrithub-cl-contributions

And yes, definitely agreed that we can add the new API as a separate contribution.

mvdan:
Imported as https://review.gerrithub.io/c/cue-lang/cue/+/1173735. I also added a brief note to the commit message clarifying what changes the user might see from this switch.

PR #2729, opened 2023-12-08
fix #2720 yaml encoding of hex strings

### LINKED ISSUE #2720: Bug when outputting hex strings to yaml

<pre>
cue version v0.6.0

go version go1.21.2
      -buildmode exe
       -compiler gc
  DefaultGODEBUG panicnil=1
     CGO_ENABLED 1
          GOARCH amd64
            GOOS linux
         GOAMD64 v1
</pre>

yes

### What did you do?
```
example:  "0x12345678901234567890"
example2: "0x123"
```

```
example: "0x12345678901234567890"
example2: "0x123"
```

### What did you see instead?
```
example: 0x12345678901234567890
example2: "0x123"
```

I'm 90% sure this is caused by this bug:
https://github.com/go-yaml/yaml/issues/847

This is very problematic in the blockchain space because ethereum addresses, block hashes and many other things look like hex strings.

vikstrous:
See #2720 for details


mvdan (commented):
Thanks for the fix! I'm not particularly familiar with internal/encoding/yaml myself, but the fix looks simple and clear enough to me.

The test case should go in the `encoding/yaml` package, for example as part of `TestYAMLValues`, since `internal/encoding/yaml` is our fork of go-yaml with all the inherited tests. If/when we eventually update our go-yaml fork or move to an entirely different YAML library, I definitely don't want to lose this regression test :)

vikstrous2:
I added the test case to encoding/yaml.

mvdan:
Thank you! Imported to Gerrit as https://review.gerrithub.io/c/cue-lang/cue/+/1173689.

I hope you don't mind that I wrote a longer commit message to add context to the git log, and I also added a note in the test case to remind ourselves that upstream hasn't fixed this yet. I also removed the internal/encoding/yaml test, as it was now redundant.

vikstrous:
Great! Thanks!

PR #2758, opened 2024-01-06
cmd/cue: fix how to trim prefix

nnnkkk7:
I modified trim process in test to be clearer.


mvdan (approved):
Thanks - I'll tweak the commit message as it's not really a fix, but sounds fine to me.

PR #2759, opened 2024-01-06
cmd/cue: fix help cmd incorrect behavior

### LINKED ISSUE #2560: cmd/cue: help cmd $cmd does not exit non-zero for unknown command

<pre>
$ cue version
cue version v0.0.0-20230821113919-220b44e973ef

go version go1.20.6
      -buildmode exe
       -compiler gc
     CGO_ENABLED 1
          GOARCH arm64
            GOOS linux
             vcs git
    vcs.revision 220b44e973eff9d41f2067499f20a12b6a0a6f29
        vcs.time 2023-08-21T11:39:19Z
    vcs.modified false
</pre>

Yes

### What did you do?

```
exec cue help cmd x
stdout 'long desc'

! exec cue help cmd doesnotexist
stderr 'unknown command .* doesnotexist'

-- x_tool.cue --
package x

import "tool/cli"

command: x: cli.Print & {
	$long: "long desc"
	text: "test"
}
```

Passing test (modulo the correct expectation on `stderr`)

### What did you see instead?

```
# known command (0.016s)
# unknown command (0.012s)
> ! exec cue help cmd doesnotexist
[stdout]
cmd executes the named command for each of the named instances.

...

Run "cue help commands" for more details on tasks and commands.

Usage:
  cue cmd <name> [inputs] [flags]
  cue cmd [command]

Available Commands:
  x

Flags:
  -t, --inject stringArray   set the value of a tagged field
  -T, --inject-vars          inject system variables in tags (default true)

Global Flags:
  -E, --all-errors   print all available errors
  -i, --ignore       proceed in the presence of errors
  -s, --simplify     simplify output
      --strict       report errors for lossy mappings
      --trace        trace computation
  -v, --verbose      print information about progress

Use "cue cmd [command] --help" for more information about a command.
FAIL: /tmp/testscript1425870910/repro.txtar/script.txtar:6: unexpected command success
```

i.e. the requesting for help for an unknown command resulted in a zero exit code.

nnnkkk7:
This fixes the issue of incorrect behavior when requesting help for an unknown command in the CUE CLI.

1. Modified the `newHelpCmd` function to handle unknown commands correctly.
2. When an unknown `cue help`  commands are entered, an error message is displayed and the program exits with a non-zero status code.
3. This ensures that the CLI behaves as expected and provides useful feedback to the user when an unknown command is entered.

 Fixes #2560

### DIFF
--- a/cmd/cue/cmd/help.go
+++ b/cmd/cue/cmd/help.go
@@ -15,6 +15,9 @@
 package cmd
 
 import (
+	"os"
+	"strings"
+
 	"github.com/spf13/cobra"
 )
 
@@ -34,7 +37,7 @@ func newHelpCmd(c *Command) *cobra.Command {
 		Long: `Help provides help for any command in the application.
 Simply type ` + c.Name() + ` help [path to command] for full details.`,
 		Run: func(_ *cobra.Command, args []string) {
-			cmd, _, e := c.Root().Find(args)
+			cmd, rests, e := c.Root().Find(args)
 			if len(args) > 0 && args[0] == "cmd" {
 				// args is one of:
 				//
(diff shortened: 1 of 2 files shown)


mvdan:
This would need a test; see the other `cue help` test scripts.

Also, I suspect this will break existing tests. Did you run them?

mvdan (changes requested):
See the above

nnnkkk7:
thanks! i fixed it!

mvdan:
> This would need a test; see the other `cue help` test scripts.

See the above - if you are fixing a bug, a test should be added. See https://github.com/cue-lang/cue/blob/master/CONTRIBUTING.md

nnnkkk7:
@mvdan 
I missed it. Sorry, I added the test.
[add test for os.Exit(1)](https://github.com/cue-lang/cue/pull/2759/commits/ecc3ffa3541d07afded0431098160cfa79f1d40b)

mvdan:
Thanks! No need to write a Go test and mock os.Exit for this; we already have integration tests for cmd/go at cmd/cue/cmd/testdata/script/help.txtar and sibling files like help_cmd.txtar, for example.

Also, is there a particular reason that you swapped Printf and CheckErr for PrintErrf and os.Exit? As far as I can tell, the fix here is the addition of `|| len(rests) != 0`, so the other bits of the PR shouldn't be necessary.

mvdan:
Did the above, as well as added more test cases, in this Gerrit patch: https://review.gerrithub.io/c/cue-lang/cue/+/1194247

Hope that's okay with you, and please let me know if you spot any issues there.

PR #2771, opened 2024-01-20
internal/cmd/cue-ast-print: fix panic when printing the ast tree

rudifa:
Panic was observed when running cue-ast-print 2567.slice.1.cue or similar cue samples.

This commit fixes the problem by moving the test `if !v.IsValid(){...}` a few lines up, to prevent referencing a nil pointer.

The script issue-panic.txtar verifies that cue-ast-print works after this fix.

Running the script with a build of cue-ast-print before the fix demonstrates the panic about half-way into the printing.

### DIFF
--- a/internal/cmd/cue-ast-print/main.go
+++ b/internal/cmd/cue-ast-print/main.go
@@ -75,15 +75,16 @@ func (d *debugPrinter) value(v reflect.Value) {
 	if v.Kind() == reflect.Interface {
 		v = v.Elem()
 	}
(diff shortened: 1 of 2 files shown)


mvdan:
Thanks! Importing as https://review.gerrithub.io/c/cue-lang/cue/+/1194721.

rudifa:
Happy to help!

Le mar. 14 mai 2024 à 17:23, Daniel Martí ***@***.***> a
écrit :

PR #2883, opened 2024-02-28
.pre-commit-hooks: Added pre-commit-hooks

### LINKED ISSUE #2834: project: pre-commit hooks

I would love to use cue to validate some yamls everytime I make a commit by the use of [pre-commit](https://pre-commit.com/) hooks. At the moment, it is possible to run cue commands from a pre-commit hook but that requires cue to be already installed outside the pre-commit config. 

It would be lovely if the cue repo could offer pre-commit hooks and we can manage everything by just having a cue entry in the pre-commit config. 

One option is to call cue commands by setting up a local script that pre-commit has to run. The disadvantage is that cue would need to already be installed on the machine pre-commit runs. It would be hard to adopt since that would require other repository stakeholders to install an additional dependency manually.

gauthsvenkat:
This commit adds a pre-commit-hook for common cue (sub) commands

At the moment, only `cue vet` and `cue fmt` are included.

Fixes #2834


gauthsvenkat:
This is by no means a complete pull request. I need advice on 2 things.

1) What are sensible defaults for the hooks?  
2) Which cue commands make sense to be included in the pre-commit hook? For eg. cue export would be a good candidate. 
And of course, what are the sensible defaults for those commands?

Furthermore, I also plan on adding the option to use the docker version of cue with pre-commit.

gauthsvenkat:
Closing PR along with #2834 .

PR #2984, opened 2024-03-23
cmd/cue: get go: support time.Duration

### LINKED ISSUE #2969: `cue get go` renders time.Duration with invalid CUE

<pre>
$ cue version
cue version v0.8.0

go version go1.22.1
      -buildmode exe
       -compiler gc
        -ldflags -extldflags -static
  DefaultGODEBUG httplaxcontentlength=1,httpmuxgo121=1,tls10server=1,tlsrsakex=1,tlsunsafeekm=1
     CGO_ENABLED 0
          GOARCH amd64
            GOOS linux
         GOAMD64 v1
             vcs git
    vcs.revision 7d757419377e66b0aa7bc7a6d68b6b6801214bd3
        vcs.time 2024-03-15T12:04:18Z
    vcs.modified false

</pre>

Yes

### What did you do?
Here's a txtar with a failing test, with what I believe to be the correct output.
```
# Test for go get issue with time.Duration returning invalid numerics.
#
# cue get go
exec cue get go --local ./...
cmp ./pkg1/file1_go_gen.cue ./pkg1/file1_go_gen.cue.golden

# Verify dependencies did not change
cmp go.mod go.mod.golden

-- go.mod --
module mod.test

go 1.21
-- go.mod.golden --
module mod.test

go 1.21
-- cue.mod/module.cue --
module: "mod.test"
-- pkg1/file1.go --
package pkg1

import "time"

const (
	Time1Second = time.Duration(time.Second)
)
-- pkg1/file1_go_gen.cue.golden --
// Code generated by cue get go. DO NOT EDIT.

//cue:generate cue get go mod.test/pkg1

package pkg1

import "time"

Time1Second: time.Duration & "1000000000ns"
```

The `time.#Duration` to be properly referenced as `time.Duration`.
The duration represented as a valid string, not an invalid numeric.
Also the proper unit suffixed (ns).

### What did you see instead?
A numeric for the time, and an invalid `time.#Duration` reference.

```
❯ diff pkg1/file1_go_gen.cue pkg1/file1_go_gen.cue.golden 
9c9
< #Time1Second: time.#Duration & 1000000000
---
> Time1Second: time.Duration & "1000000000ns"
```

gdvalle:
Fix `get go` behavior with `time.Duration` to output valid CUE.
Convert the underlying numeric to string, and add the unit suffix so
it's compliant with CUE's `time.Duration`.

Add `Duration` to the `time` builtin check so we render as
`time.Duration` rather than the invalid `time.#Duration`.

Fixes #2969.


cmd/cue/cmd/testdata/script/issue2969.txtar
```diff
@@ -0,0 +1,37 @@
...
+-- pkg1/file1_go_gen.cue.golden --
+// Code generated by cue get go. DO NOT EDIT.
+
+//cue:generate cue get go mod.test/pkg1
+
+package pkg1
+
+import "time"
+
+#Time1Second: time.Duration & "1000000000ns"
```
mvdan: it seems to me like this should be `1s`; perhaps we can have a bit of logic that uses larger units as long as no info is lost. for example, `3000ms` could be `3s`, but `3030ms` could remain as-is. We could also consider e.g. `3m3ms` for the latter case, but that feels like diminishing returns.

cmd/cue/cmd/get_go.go
```diff
@@ -788,8 +788,16 @@ func (e *extractor) reportDecl(x *ast.GenDecl) (a []cueast.Decl) {
 					}
 				}
 
+				typ := e.pkg.TypesInfo.TypeOf(name)
 				c := e.pkg.TypesInfo.Defs[v.Names[i]].(*types.Const)
 				sv := c.Val().ExactString()
+				switch t := typ.(type) {
+				case *types.Named:
+					if t.Obj().Pkg().Path() == "time" && t.Obj().Name() == "Duration" {
+						// time.Duration is an underlying int64, but CUE only accepts string.
```
mvdan: to be a bit more specific:

> // time.Duration is an int64 representing nanoseconds,
> // but CUE only accepts strings with units as accepted by time.ParseDuration.

cmd/cue/cmd/testdata/script/issue2969.txtar
```diff
@@ -0,0 +1,37 @@
+# Test for go get issue with time.Duration returning invalid numeric.
+#
+# cue get go
```
mvdan: remove this line? it doesn't seem to add anything

cmd/cue/cmd/get_go.go
gdvalle: done in https://github.com/cue-lang/cue/pull/2984/commits/fa8b5594062955ab4b7a28676176cd55dc273e92

cmd/cue/cmd/testdata/script/issue2969.txtar
gdvalle: I think that makes good sense for the humans. In https://github.com/cue-lang/cue/pull/2984/commits/fa8b5594062955ab4b7a28676176cd55dc273e92 I figured we could just parse the string back into a proper `time.Duration`, and use `String()` to pull it back out in a reduced form.

cmd/cue/cmd/testdata/script/issue2969.txtar
gdvalle: done in https://github.com/cue-lang/cue/commit/fa8b5594062955ab4b7a28676176cd55dc273e92

cmd/cue/cmd/testdata/script/issue2969.txtar
mvdan: Reusing Duration.String is clever :) SGTM

mvdan (approved):
Thanks! I will import into Gerrit soon.

gdvalle:
Thank you, mvdan!

I'll open these directly on Gerrit to save you some effort in the future (or if you'd appreciate me re-submitting over there, just say the word, it's no trouble).

mvdan:
If you are able to create a patch on Gerrit, yes, that would be better :) Even now after we have reviewed the patch, it saves the indirection of me being the uploader even though I'm not the author.

gdvalle:
Opened https://review.gerrithub.io/c/cue-lang/cue/+/1189718, so I'll close this now.

Issue #2035, opened 2022-10-24
CUE and WebAssembly (Wasm) 


myitcv:
This is an umbrella tracking issue for WebAssembly support in CUE. 

Details to follow, but initial exploration is linked to #2007.

4ad:
If you'd like to provide use cases and/or feedback on potential use of Wasm with CUE, please post to the discussion at https://github.com/cue-lang/cue/discussions/2045.

sdboyer:
We have a solid use case for WASM with Thema - https://github.com/grafana/thema/issues/36.

Basically, we want a Thema-specific CUE playground-ish environment, which lets us present the user with all the basic thema operations. (Essentially, the operations exposed via the `thema` CLI).

Certainly, it would make [docs like this one](https://github.com/grafana/thema/blob/d936c1921950c4517cb9da1cb994340aa9cf6990/docs/go-quickstart.md) far more powerful - instead of asking people to run code generation, try out validation, etc., on their local, they can see it directly in the browser. But it would also enable us to do things like - just one example - update tutorials we write for configuring systems on the fly/in the browser, if their config file is written in Thema.

It's important to note that we need to have some Go logic in there, not just pure CUE, because we can't express all the constraints/invariants required by Thema in native CUE, and because our code generators are written in Go.

myitcv:
For folks following along here, apologies we forgot to add notice of the [CUE + Web Assembly Focus Session Call](https://github.com/cue-lang/cue/discussions/2463). It was well advertised elsewhere, but we failed to join the dots here.

4ad:
As we discussed in several community calls, we're pausing Wasm developement until the Wasm ecosystem matures, in particular, we're waiting for the [Component Model](https://component-model.bytecodealliance.org).

AndrewSav:
Will there be an alternate way to do custom functions?

4ad:
We are open to the idea, although we don't have a concrete design in mind yet. Care to share an experience report, or a simple feature request?

yoktobit:
Would using an abstraction layer like extism help for integration of wasm? Then cue using wasm/wasi/wasipx wouldn't depend too much on their versions but more on the implementation in extism. https://github.com/extism/go-sdk

keeganmccallum:
In case useful: https://github.com/dclareio/cue-wasm

myitcv:
[...]

Issue #350, opened 2021-07-03
cue: Enable builtins to be registered by users


cueckoo:
_Originally opened by @verdverm in https://github.com/cuelang/cue/issues/350_

As a user, I would like to register my own builtins to be used in Cue.

- Expose a `RegisterBuiltin` function which allows adding builtins to the predefined map
- add the builtins to the runtime (or something else)? This would allow users to have multiple runtimes and different builtins per.
- Enable new stdlib functions to be developed, tested, and used prior to modifying or extending  the actual list of builtins 
- Collect and document how to decide and register as pure vs tool. Alert to potential issues with non-pure builtins in pure cue, can shoot yourself in the foot.

cueckoo:
_Original reply by @shykes in https://github.com/cuelang/cue/issues/350#issuecomment-619291671_

Doesn't this create the risk of fragmenting the language? If I load a file with the extension `.cue`, I have certain expectations of available builtins and their behavior. Even more so if we allow builtins to not be pure (no real way to avoid it without extreme sandboxing which is another can of worms). 

If custom builtins become widely adopted, it kind of changes the role of Cue from a language to a family of similar but incompatible languages. This in turn could hurt adoption of Cue, and the benefits of using it.

What are your thoughts on this?

cueckoo:
_Original reply by @verdverm in https://github.com/cuelang/cue/issues/350#issuecomment-619322818_

Super good question. @mpvl, would love to hear your thoughts

There are definitely a few that people have been asking for. I have a strong desire for the main functions in [structural](https://github.com/hofstadter-io/structural). I'm about to implement them in Go, along the lines of the update / mutation prototype we got working in slack today.

#348 is the other half of this, to have a place to talk about what builtins Cue should have. There are also the annotations for tooling to work with, and a lot should be possible there. Based on your perspective @shykes, what @mpvl thinks, this PR could go to the chopping block pretty quick.

Another thing that came out of the prototype is the difficulty several of us experienced constructing or mutation Cue values in our Golang based tools. If we improve that, so should the experience around building tools around Cue. At the same time, many of us would like to push as much into Cue as we can. It does seem more tedious to switch back and forth, but then again, maybe we can fix that in the Go API area.

---

[...]

PR #3200, opened 2024-06-05
pkg/text/template: Add FuncMap support in text/templates

thatInfrastructureGuy:
* This attempts to _partially_ resolve: https://github.com/cue-lang/cue/discussions/2498 
* Does not change CUE API for text/templates
* This only enables users who use Go API, to create user-defined Go functions via text/template
* Since text/template only returns `string`, users will need to use `strconv` library for other formats.

### Usage

example.cue
```cue
package gen

import (
	"text/template"
	"strconv"
)

isadult: bool

#person: {
	Name: "John"
}

#templatetext: "{{ .Name }} is of age {{ GetAge .Name }}."
#checkadult:   "{{ IsAdult .Name }}"

message: template.Execute(#templatetext, #person)
isadult: strconv.ParseBool(template.Execute(#checkadult, #person))
```

main.go
```go
package main

import (
	"fmt"
	"log"

	"cuelang.org/go/cue/load"
	"cuelang.org/go/encoding/yaml"
	"cuelang.org/go/pkg/text/template"
)

var person = map[string]int{
	"John": 25,
	"Jane": 30,
}

func GetAge(name string) int {
	return person[name]
}

func main() {
	ctx := cuecontext.New()
	template.AddFuncMap(template.FuncMap{
		"GetAge": GetAge,
		"IsAdult": func(name string) bool {
			if GetAge(name) > 18 {
				return true
			}
			return false
		},
	})
	// Load the CUE file
	instances := load.Instances([]string{"./example.cue"}, nil)
	v := ctx.BuildInstance(instances[0])
	if v.Err() != nil {
		log.Fatalf("Failed to load CUE file: %v", v.Err())
	}

	// Export as YAML
	data, err := yaml.Encode(v)
	if err != nil {
		log.Fatalf("Failed to marshal as YAML: %v", err)
	}

	// Output YAML to stdout
	fmt.Println(string(data))
}
```

### DIFF
--- a/pkg/text/template/manual.go
+++ b/pkg/text/template/manual.go
@@ -41,7 +41,9 @@ func AddFuncMap(funcMap FuncMap) {
 
 // Execute executes a Go-style template.
 func Execute(templ string, data cue.Value) (string, error) {
+	muFM.Lock()
 	t, err := template.New("").Funcs(fm).Parse(templ)
+	muFM.Unlock()
 	if err != nil {
 		return "", err
 	}
--- a/pkg/text/template/manual.go
+++ b/pkg/text/template/manual.go
@@ -15,15 +15,33 @@
 package template
 
 import (
+	"maps"
 	"strings"
+	"sync"
 	"text/template"
 
 	"cuelang.org/go/cue"
 )
 
+type FuncMap template.FuncMap
+
+var (
+	fm   = make(template.FuncMap)
+	muFM = sync.Mutex{}
+)
+
+// AddFuncMap copies funcMap provided by the user to global map.
+// If AddFuncMap is called multiple times with a key already present in global map,
+// the value in the global map will be overwritten by the latest value associated
+func AddFuncMap(funcMap FuncMap) {
+	muFM.Lock()
+	defer muFM.Unlock()
+	maps.Copy(fm, funcMap)
+}
+
 // Execute executes a Go-style template.
 func Execute(templ string, data cue.Value) (string, error) {
-	t, err := template.New("").Parse(templ)
+	t, err := template.New("").Funcs(fm).Parse(templ)
 	if err != nil {
 		return "", err
 	}


pkg/text/template/manual.go
```diff
@@ -15,15 +15,33 @@
...
+func AddFuncMap(funcMap FuncMap) {
+	muFM.Lock()
+	defer muFM.Unlock()
+	maps.Copy(fm, funcMap)
+}
+
 // Execute executes a Go-style template.
 func Execute(templ string, data cue.Value) (string, error) {
-	t, err := template.New("").Parse(templ)
+	t, err := template.New("").Funcs(fm).Parse(templ)
```
verdverm, line 45: This should be protected by the mutex

Otherwise, if Execute is called concurrently with AddFuncMap, there will be a race

thatInfrastructureGuy: Thanks @verdverm. Updated
https://github.com/cue-lang/cue/blob/1c3c202cde93828510abf15c0498605a4dcc6d43/pkg/text/template/manual.go#L44-L46

thatInfrastructureGuy:
Related conversation in [slack](https://cuelang.slack.com/archives/CLT3ULF6C/p1715876190049979?thread_ts=1715876190.049979&cid=CLT3ULF6C)

gotwarlost:
Drive-by comment: since this only benefits Go users, I wonder if this should not just be a new method that accepts a pre-created template object instead of registering a func map, dealing with locks etc.  i.e. 

```
func ExecuteTemplate(t *template, data cue.Value) (string, error)
```

I'm probably missing some nuance here....

verdverm:
This enables you to add functions to the template call from CUE values.

Arguably, this could create a situation where there are multiple dialects of CUE, much like user defined functions. (i.e. The CUE given as example will not be usable with the `cue` CLI. Another thing this can enable is access to the external world, breaking CUE's hermetic nature.

Maybe it is better to use the WASM integration for this? (create a WASM function with your extra helpers that also calls execute)

As this creates "sub-dialects", I would generally be against this, especially with the WASM option beginning to materialize

Relevant discussion about fragmenting the ecosystem: https://github.com/cue-lang/cue/issues/350

thatInfrastructureGuy:
@verdverm Thanks again for your review!

> The CUE given as example will not be usable with the cue CLI.

Agreed. I was hoping not to change any API for CUE users. Since, [WASM development is paused](https://github.com/cue-lang/cue/issues/2035#issuecomment-2112892919) this can be a stopgap solution until then.

>  Another thing this can enable is access to the external world, breaking CUE's hermetic nature.

Agreed again :) This is not meant for general CUE usage. It only serves consumers consuming via API. In that case, even via [`FillPath`](https://pkg.go.dev/cuelang.org/go/cue#Value.FillPath) we break the hermetic nature.

I do :100: agree about having sub-dialects. I am personally content having a soft-fork with CI pipeline which adds this patch (until WASM support). I just brought this PR up to see if the community benefits from it.

thatInfrastructureGuy:
@gotwarlost Thanks for the review. I wanted to do smallest possible change which does not affect CUE API. With this PR, I am trying to get feedback if something like this makes sense. Your comment makes sense to me. I do not have the knowledge right now to execute this correctly. But, I can dig deeper.

> pre-created template object 

This might not work for allowing any user-defined function.

myitcv:
Thanks all for the discussion here, and @thatInfrastructureGuy thank you for helping to move this conversation forward. 

Per my "bump" response in https://github.com/cue-lang/cue/discussions/2498#discussioncomment-9734094, we are actively looking at an alternative to WASM to fill the gap left by pausing work in that space. 

It's not clear to me that we should make this change in the `text/template` proposed here because of the reasons outlined above:

* It's limited to users of the Go API
* It's limited to use in `text/template` as opposed to extending the builtins

We will continue our investigation of a more general approach. 

My suggestion would be that we shift conversation from this PR back to #2498 and related issues/discussions. @thatInfrastructureGuy would that work for you?

My further suggestion would be that we close this PR, given the reasoning above.

thatInfrastructureGuy:
> My suggestion would be that we shift conversation from this PR back to #2498 and related issues/discussions. @thatInfrastructureGuy would that work for you?

Thanks @myitcv for the review. That makes sense to me.

PR #3569, opened 2024-11-10
encoding/jsonschema: allow "date" format for OpenAPI

### LINKED ISSUE #3568: encoding/jsonschema: "format" validation is too strict
I was trying to import an OpenAPI spec and it failed because of `"format": "date"` being used. OpenAPI 3.0 explicitly supports the "date" format for strings, so that seems like a bug to me.

I looked into where the error is coming from to submit a fix and to me it seems like the validation on the "format" field seems too strict. Even though JSON Schema and OpenAPI define some known formats, they both describe the "format" field as allowing any value, so I don't think a non-standard value should cause the import to fail.

- [OpenAPI 3.0.0 Data Types](https://spec.openapis.org/oas/v3.0.0.html#data-types)
- [JSON Schema "Format" reference](https://json-schema.org/understanding-json-schema/reference/string#format)

From the master branch. (commit af459f14b88efc59bee88f3b3a0f8b1b37dbfa66)

<pre>
$ cue version

cue version v0.0.0-20241106162726-af459f14b88e
</pre>

No, only on the master branch.

### What did you do?

```sh
cue import openapi openapi.json
```

No output or errors and an `openapi.cue` file is created.

### What did you see instead?

The command fails with the following error being logged many times:

```
format "date" is not recognized in schema version OpenAPI 3.0:
    ./openapi.json:<line>:<pos>
```

justenstall:
Fixes #3568

### DIFF
--- a/encoding/jsonschema/constraints_format.go
+++ b/encoding/jsonschema/constraints_format.go
@@ -31,7 +31,7 @@ var formatFuncs = sync.OnceValue(func() map[string]formatFuncInfo {
 		"binary":                {openAPI, formatTODO},
 		"byte":                  {openAPI, formatTODO},
 		"data":                  {openAPI, formatTODO},
-		"date":                  {vfrom(VersionDraft7), formatTODO},
+		"date":                  {vfrom(VersionDraft7) | openAPI, formatTODO},
 		"date-time":             {allVersions | openAPI, formatDateTime},
 		"double":                {openAPI, formatTODO},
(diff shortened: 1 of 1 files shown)


rogpeppe:
Thanks for this contribution! However, I've chosen to go with a change that actually adds proper
support for the `date` format instead.

https://review.gerrithub.io/c/cue-lang/cue/+/1203871

Hopefully that should fulfil your needs here.

justenstall:
@rogpeppe Yep, that works!

PR #3740, opened 2025-02-10
internal/encoding: don't interpret file as UTF8 with binary encoding

nichtsundniemand:
Using, for example, `@embed(type=binary)` CUE would interpret the file's contents as UTF8. Since not all byte-sequences are valid UTF8, some will not be interpretable. Bytes sequences that fail to be properly interpreted will not raise an evaluation error but instead yield erroneus bytes to the evaluator.

This change makes it so the file's contents will be passed to the evaluator verbatim.

### DIFF
--- a/cmd/cue/cmd/testdata/script/embed.txtar
+++ b/cmd/cue/cmd/testdata/script/embed.txtar
@@ -4,6 +4,16 @@ env CUE_EXPERIMENT=embed=0
 cmp stderr out/noembed
 env CUE_EXPERIMENT=
 
+# Create test files for type=binary. txtar can only hold valid UTF8 - we need to test arbitrary bytes as input so we use quoted strings for storage
+# Text file, encoded in UTF8-8 with leading BOM
+# BOM should be stripped by faulty binary handling
+strconv-unquote '"\xef\xbb\xbfHello, 世界"'
+cp stdout test-utf8.binary
+# A byte sequence - should be invalid UTF-8
+# Each byte should be replaced by a unicode replacement character due to faulty binary handling
+strconv-unquote "\xf0\xf1\xf2\xf3\xf4\xf5\xf6\xf7"
+cp stdout test-nonutf8.binary
+
 exec cue eval
 cmp stdout out/eval
 
@@ -38,6 +48,12 @@ d: _ @embed(glob="x/*.yaml") // merge into the same map
 f: _ @embed(file="openapi.json", type=openapi)
 
 g: _ @embed(file="openapi.json") // test no auto mode!
+
+h: _ @embed(file="test-utf8.binary", type=binary)
+
+i: _ @embed(file="test-utf8.binary", type=text)
+
+j: _ @embed(file="test-nonutf8.binary", type=binary)
 
 special: {
 	// These are all valid.
@@ -213,6 +229,18 @@ g: {
         }
     }
 }
+h: '''
+    Hello, 世界
+
+    '''
+i: """
+    Hello, 世界
+
+    """
+j: '''
+    ��������
+
+    '''
 special: {
     underscoreFile: {
         z: 45
@@ -273,6 +301,18 @@ g: {
 		}
 	}
 }
+h: '''
+	Hello, 世界
+
+	'''
+i: """
+	Hello, 世界
+
+	"""
+j: '''
+	��������
+
+	'''
 special: {
 	// These are all valid.
 	underscoreFile: z:  45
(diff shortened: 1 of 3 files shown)


mvdan:
SGTM but this would need a test; see my longer response at https://github.com/cue-lang/cue/issues/3741.

nichtsundniemand:
@mvdan OK, so I took you idea from https://github.com/cue-lang/cue/issues/3741#issuecomment-2651179550

There appears to be some issue with testscript, maybe, because I was trying `strconv-unquote "Hello, world!"` and I would get some syntax-errors. If the argument-strings don't contain any spaces the errors disappear :/

In my case it's not so important since I don't need to put spaces, just something I noticed.

The valid-UTF8 file contains a BOM at the start since the UTF8 decoder also does BOM stripping and I wanted to show that `type=text` will correctly strip that while `type=binary` preserves it.

The non-UTF8-file just contains some invalid byte sequence(s). Running that test against `master` without the fix produces a byteslice with 8 of those Unicode-Replacement chars.

mvdan:
> There appears to be some issue with testscript, maybe, because I was trying `strconv-unquote "Hello, world!"` and I would get some syntax-errors. If the argument-strings don't contain any spaces the errors disappear :/

Ah, that is because testscript is not POSIX shell. It only has single quotes, not double quotes, and they behave differently than POSIX shell too. See https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript.

So you can either do `strconv-unquote "foo"` or `strconv-unquote '"foo bar baz"'`. My apologies that my original example in https://github.com/cue-lang/cue/issues/3741#issuecomment-2651179550 omitted to note this pitfall.

cmd/cue/cmd/testdata/script/embed.txtar
```diff
@@ -4,6 +4,13 @@ env CUE_EXPERIMENT=embed=0
 cmp stderr out/noembed
 env CUE_EXPERIMENT=
 
+# Create test files for type=binary. txtar can only hold valid UTF8 and strips any BOM - we need to test arbitrary bytes
+# as input so we use quoted strings for storage
```
mvdan: please document what each of the two files contains, i.e. the first one is utf8 with a BOM, and the second one isn't valid utf8 at all.

mvdan (approved):
Thanks!

mvdan:
Do you want to merge this as one commit, or as three?

If you want to merge as one commit, that's easy; please squash all three commits into one, making sure that the resulting commit has a good commit message, probably combining the three you already wrote.

If you want two or three commits, then the tests must pass at each commit for us to be able to merge this. Right now, if I checkout the second commit, the tests fail, as they expect the fix from the next commit.

Two commits rather than one is actually nice, where the first commit shows the _wrong behavior_ in the test, and the fix changes the test to expect the fixed behavior. That way, both commits pass the tests, and you can see the effect of the fix - and also verify that the test is actually testing what it's meant to.

nichtsundniemand:
> Two commits rather than one is actually nice, where the first commit shows the _wrong behavior_ in the test, and the fix changes the test to expect the fixed behavior. That way, both commits pass the tests, and you can see the effect of the fix - and also verify that the test is actually testing what it's meant to.

Okay, i will rework the test like this because I was thinking a similar thing about showing that the fix works.

cmd/cue/cmd/testdata/script/embed.txtar
nichtsundniemand: Done

As I was mistaken about BOM-stripping by txtar, that part of the comment has been removed. Although it's technically possible to losslessly pack a text-file with leading BOM in a txtar the "valid-UTF8-with-BOM" testfile is still created using strconv-unquote.

I believe it is much easier to miss a non-printing-character, like the BOM, in a git diff or while editing the testdata. Writing the BOM-bytes out makes it more visible.

mvdan:
Please ignore the CI failure; that panic looks unrelated and it could be a Go bug: https://github.com/golang/go/issues/71451

mvdan:
Please let me know when you are done fixing up the commits. Ideally we end up with two commits here - one adding the test and strconv-unquote, and another doing the fix and updating the test.

nichtsundniemand:
@mvdan Thanks for the follow-up. I was pretty much done last week and that test-failure tripped me up...

I wouldn't mind three commits but according to your input I squashed the first two.

IMO there's nothing left to do for me right now :)

mvdan (approved):
Thank you!

mvdan:
Imported the two commits to Gerrit here: https://review.gerrithub.io/c/cue-lang/cue/+/1209313

I took the liberty of slightly rewriting your test comments and commit messages, I hope that's okay - mainly to add the "fixes" lines referencing the issue, and to shorten the texts a bit. The code and test logic is unchanged.

mvdan:
Also, please feel free to send patches to Gerrit directly in the future, as that way we can do the review and merging directly there. GitHub is only a mirror in terms of git; we don't do any reviews or merges here.

nichtsundniemand:
Thanks alot! I appreciate the edits.

I will check out the GerritHub workflow for the future.

Does this PR require any further action or will it be closed automatically?

mvdan:
All good, I'll take it from here.

PR #3808, opened 2025-03-11
encoding/koala: add the Decoder for the koala XML encoding

matthew-sladescu:
This PR adds a Decoder for an XML encoding for CUE called koala, as described in the proposal [here](https://github.com/cue-lang/cue/discussions/3776).


mvdan:
For those who might have seen or followed this PR, it has moved to Gerrit: https://review.gerrithub.io/c/cue-lang/cue/+/1211362

Issue #3142, opened 2024-05-13
ip arithmetic


lgfausak:
i create a lot of json to do configuration. typically my configurations contain ip addresses. i am new to cue, i've spent the day doing experiments and reading through all of the docs / tutorials / etc.  If I wanted to use cue to create an ip address based upon a cidr + offset is that possible?  something like "192.168.0.0/24" + 3 == 192.168.0.3

NoamTD:
it's a bit convoluted, but would something like this work for you?

```cue
import "net"

#IncIP: {
    In!: net.IPv4
    Offset!: int

    _incremented: In[:3] + [In[len(In)-1]+Offset]

    Out: net.IPString(_incremented)
}

sourceIp: net.ToIP4("192.168.0.0")
withOffset: (#IncIP & {Offset: 1, In: sourceIp}).Out
```

output:
```
sourceIp: [192, 168, 0, 0]
withOffset: "192.168.0.1"
```

slewiskelly:
> it's a bit convoluted, but would something like this work for you?

This will fail if exceeding an octet (255).

I'm not certain that this is totally correct, but:

```cue
import (
	"math"
	"net"
	"strconv"
)

ip: (#IPAdd & {ip: "192.168.0.0", i: 256}).out

#IPAdd: {
	ip: string
	i:  uint

	out: "\(math.Floor(_out/16777216)).\(math.Floor(_out/65536) mod 256).\(math.Floor(_out/256) mod 256).\(_out mod 256)"

	_ip:  net.ToIP4(ip)
	_out: (_ip[0]*16777216 + _ip[1]*65536 + _ip[2]*256 + _ip[3]) + i
}
```

This is easier to implement in Go, but I'm not sure if something like this would be added to the standard library, though.

myitcv:
> but I'm not sure if something like this would be added to the standard library

This seems like a reasonable thing to support. Not least because of the problems that arise when trying to "do this by hand" as demonstrated in #3787. But generally because it seems useful enough to have cropped up here more than once. 

Question: is it generally the case that a plain integer value is added? Asked another way, what should the API of such a function look like? Any good examples from the Go standard library or elsewhere?

takonomura:
@myitcv ​Thank you for considering support for IP address arithmetic in the standard library.

> Question: is it generally the case that a plain integer value is added? Asked another way, what should the API of such a function look like? Any good examples from the Go standard library or elsewhere?

[...]

Issue #3787, opened 2025-02-27
eval: loss of Precision in Arithmetic Operations on 35-Digit Integers


takonomura:
### What version of CUE are you using (`cue version`)?

<pre>
$ cue version
cue version v0.12.0

go version go1.23.5
      -buildmode exe
       -compiler gc
       -trimpath true
  DefaultGODEBUG asynctimerchan=1,gotypesalias=0,httpservecontentkeepheaders=1,tls3des=1,tlskyber=0,x509keypairleaf=0,x509negativeserial=1
     CGO_ENABLED 1
          GOARCH arm64
            GOOS darwin
         GOARM64 v8.0
cue.lang.version v0.12.0
</pre>

### Does this issue reproduce with the latest stable release?

Yes

### What did you do?

```
exec cue eval
cmp stdout stdout.golden

-- x.cue --
package x

a: 10000000000000000000000000000000001
b: 10000000000000000000000000000000000 + 1
c: 10000000000000000000000000000000002 - 1

// 34-digit integer works
x: 9999999999999999999999999999999998
y: 9999999999999999999999999999999997 + 1
z: 9999999999999999999999999999999999 - 1
-- stdout.golden --
a: 10000000000000000000000000000000001
b: 10000000000000000000000000000000001
c: 10000000000000000000000000000000001
x: 9999999999999999999999999999999998
y: 9999999999999999999999999999999998
z: 9999999999999999999999999999999998
```

### What did you expect to see?

I expected the test to pass.
According to https://cuelang.org/docs/reference/spec/#numeric-values, I expected arithmetic operations on integer values to be performed with at least 256-bit precision.

### What did you see instead?

```
$ testscript test.txtar
> exec cue eval
[stdout]
a: 10000000000000000000000000000000001
b: 1.000000000000000000000000000000000e+34
c: 1.000000000000000000000000000000000e+34
x: 9999999999999999999999999999999998
y: 9999999999999999999999999999999998
z: 9999999999999999999999999999999998
> cmp stdout stdout.golden
diff stdout stdout.golden
--- stdout
+++ stdout.golden
@@ -1,6 +1,6 @@
 a: 10000000000000000000000000000000001
-b: 1.000000000000000000000000000000000e+34
-c: 1.000000000000000000000000000000000e+34
+b: 10000000000000000000000000000000001
+c: 10000000000000000000000000000000001
 x: 9999999999999999999999999999999998
 y: 9999999999999999999999999999999998
[...]

PR #3815, opened 2025-03-13
pkg/net: add ip arithmetic functions

takonomura:
To address #3142, this PR introduces two functions for performing arithmetic operations on IP addresses and CIDR networks. These functions may be useful for use cases such as IP address management in CUE.

### `net.AddIP`

`AddIP` adds a numerical offset to a given IP address.

This function is useful for automatically generating IP addresses for a large number of hosts, as demonstrated in the example below:

```cue
let networks = [
        "192.168.0.0/24",
        "2001:db8::/64",
]
for i in list.Range(1, 10, 1) {
        "testhost-\(i)": {
                interfaces: eth0: [for network in networks {net.AddIP(network, i)}]
        }
}
```

### `net.AddIPCIDR`

`AddIPCIDR` adds a numerical offset to a given CIDR subnet.

This function is useful for generating network prefixes based on specific rules, as demonstrated in the example below:

```cue
#VLAN: {
	id:     >=1 & <=4096
	prefix: net.AddIPCIDR("10.0.0.0/24", id)
}

vlanA: #VLAN & {id: 1}    // prefix: 10.0.1.0/24
vlanB: #VLAN & {id: 2}    // prefix: 10.0.2.0/24
vlanC: #VLAN & {id: 3}    // prefix: 10.0.3.0/24
vlanX: #VLAN & {id: 1000} // prefix: 10.3.232.0/24
```

### DIFF
--- a/pkg/net/ip.go
+++ b/pkg/net/ip.go
@@ -16,7 +16,9 @@
 package net
 
 import (
+	"errors"
 	"fmt"
+	"math/big"
 	"net/netip"
 
 	"cuelang.org/go/cue"
@@ -241,4 +243,64 @@ func IPString(ip cue.Value) (string, error) {
 		return "", fmt.Errorf("invalid IP %q", ip)
 	}
 	return ipdata.String(), nil
+}
+
+func netIPAdd(addr netip.Addr, offset *big.Int) (netip.Addr, error) {
+	i := big.NewInt(0).SetBytes(addr.AsSlice())
+	i = i.Add(i, offset)
+
+	if i.Sign() < 0 {
+		return netip.Addr{}, errors.New("IP address arithmetic resulted in out-of-range address (underflow)")
+	}
+
+	b := i.Bytes()
+	size := addr.BitLen() / 8
+
+	if len(b) > size {
+		return netip.Addr{}, errors.New("IP address arithmetic resulted in out-of-range address (overflow)")
+	}
+
+	if len(b) < size {
+		b = append(make([]byte, size-len(b), size), b...)
+	}
+	addr, _ = netip.AddrFromSlice(b)
+	return addr, nil
+}
+
+// AddIP adds a numerical offset to a given IP address.
+// The address can be provided as a string, byte array, or CIDR subnet notation.
+// It returns the resulting IP address or CIDR subnet notation as a string.
+func AddIP(ip cue.Value, offset *big.Int) (string, error) {
+	prefix, err := netGetIPCIDR(ip)
+	if err == nil {
+		addr, err := netIPAdd(prefix.Addr(), offset)
+		if err != nil {
+			return "", err
+		}
+		return netip.PrefixFrom(addr, prefix.Bits()).String(), nil
+	}
+	ipdata := netGetIP(ip)
+	if !ipdata.IsValid() {
+		return "", fmt.Errorf("invalid IP %q", ip)
+	}
+	addr, err := netIPAdd(ipdata, offset)
+	if err != nil {
+		return "", err
+	}
+	return addr.String(), nil
+}
+
+// AddIPCIDR adds a numerical offset to a given CIDR subnet
+// string, returning a CIDR string.
+func AddIPCIDR(ip cue.Value, offset *big.Int) (string, error) {
+	prefix, err := netGetIPCIDR(ip)
+	if err != nil {
+		return "", err
+	}
+	shifted := big.NewInt(0).Lsh(offset, (uint)(prefix.Addr().BitLen()-prefix.Bits()))
+	addr, err := netIPAdd(prefix.Addr(), shifted)
+	if err != nil {
+		return "", err
+	}
+	return netip.PrefixFrom(addr, prefix.Bits()).String(), nil
 }
--- a/pkg/net/testdata/gen.txtar
+++ b/pkg/net/testdata/gen.txtar
@@ -37,6 +37,19 @@ t31: net.URL & "https://foo.com/bar"
 t32: net.AbsURL & "/foo/bar"
 t33: net.AbsURL & "https://foo.com/bar"
 t34: net.AbsURL & "%"
+t35: net.AddIP("127.0.0.1", 1)
+t36: net.AddIP("127.0.0.1/8", 2)
+t37: net.AddIP("2001:db8::", 1)
+t38: net.AddIP("2001:db8::/64", 2)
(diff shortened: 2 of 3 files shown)


myitcv:
@takonomura thanks for raising this. We are extremely busy in the run up to [KubeCon](https://events.linuxfoundation.org/kubecon-cloudnativecon-europe/register/) so please bear with us!

takonomura:
Hi @myitcv, any updates on this PR? Let me know if I can do anything. Thanks!

takonomura:
Hi @myitcv, just wanted to follow up on this PR again.

If there are any concerns about the implementation or if you'd prefer a different approach, I'm happy to discuss alternatives. Otherwise, please let me know if there's anything I can do to help move this forward.

myitcv:
Thanks for the reminder @takonomura - I have asked @rogpeppe for thoughts on the API here.

pkg/net/ip.go
```diff
@@ -242,3 +244,62 @@ func IPString(ip cue.Value) (string, error) {
 	}
 	return ipdata.String(), nil
 }
+
+func netIPAddOffset(addr netip.Addr, offset *big.Int) (netip.Addr, error) {
+	i := big.NewInt(0).SetBytes(addr.AsSlice())
+	i = i.Add(i, offset)
+
+	if i.Sign() < 0 {
+		return netip.Addr{}, errors.New("underflowed")
```
rogpeppe: Maybe be a bit clearer with the error?

```suggestion
		return netip.Addr{}, errors.New("IP address arithmetic resulted in out-of-range address (underflow)")
```
(similarly below)

pkg/net/ip.go
```diff
@@ -242,3 +244,62 @@ func IPString(ip cue.Value) (string, error) {
...
+
+	b := i.Bytes()
+	size := addr.BitLen() / 8
+
+	if len(b) > size {
+		return netip.Addr{}, errors.New("overflowed")
+	}
+
+	if len(b) < size {
+		b = append(make([]byte, size-len(b)), b...)
```
rogpeppe: It's easy to avoid an extra allocation here by making the slice the correct size up front:

```suggestion
		b = append(make([]byte, size-len(b), size), b...)
```

pkg/net/ip.go
```diff
@@ -242,3 +244,62 @@ func IPString(ip cue.Value) (string, error) {
...
+	}
+	addr, _ = netip.AddrFromSlice(b)
+	return addr, nil
+}
+
+// AddIPOffset adds a numerical offset to a given IP address.
+// The address can be provided as a string, byte array, or CIDR subnet notation.
```
rogpeppe, line 271: It seems a bit odd that we can provide a CIDR string here. What's the use case for that?
For example, `net.AddIPOffset("100.0.0.0/8", 1)` returns `"100.0.0.1/8"` but that's not generally useful, as all the bits after the prefix are conventionally zero and ignored.

pkg/net/ip.go
```diff
@@ -242,3 +244,62 @@ func IPString(ip cue.Value) (string, error) {
...
+		return "", fmt.Errorf("invalid IP %q", ip)
+	}
+	addr, err := netIPAddOffset(ipdata, offset)
+	if err != nil {
+		return "", err
+	}
+	return addr.String(), nil
+}
+
+// AddIPCIDROffset adds a numerical offset to a given CIDR subnet.
```
rogpeppe:
```suggestion
// AddIPCIDROffset adds a numerical offset to a given CIDR subnet
// string, returning a CIDR string.
```

pkg/net/ip.go
```diff
@@ -242,3 +244,62 @@ func IPString(ip cue.Value) (string, error) {
...
+func AddIPCIDROffset(ip cue.Value, offset *big.Int) (string, error) {
+	prefix, err := netGetIPCIDR(ip)
+	if err != nil {
+		return "", err
+	}
+	offset.Lsh(offset, (uint)(prefix.Addr().BitLen()-prefix.Bits()))
```
rogpeppe: This side-effects the offset argument, which doesn't seems quite right from a hygiene standpoint, although it might be OK in practice, given that the `*big.Int` is probably created just for the purposes of the call. Still, I think I'd prefer to avoid doing it to be defensive and save potential nasty surprises down the line if we end up caching `*big.Int` values.

pkg/net/testdata/gen.txtar
```diff
@@ -37,6 +37,13 @@ t31: net.URL & "https://foo.com/bar"
 t32: net.AbsURL & "/foo/bar"
 t33: net.AbsURL & "https://foo.com/bar"
 t34: net.AbsURL & "%"
+t35: net.AddIPOffset("127.0.0.1", 1)
```
rogpeppe: I'd like to see test cases for the error cases too, please, including overflow, underflow and invalid inputs.

rogpeppe (commented):
I think that having this functionality available is a good thing. I just wonder if we'd maybe be better off adding something a bit more general: for example, a way to convert IP / CIDR addresses to bits or a number, and vice versa, and then we'd gain more than just this specific functionality. That said, the functions proposed here would still be more convenient to use even if we had something more general, so in general I'm in favour of this change.

One thing: I think the `Offset` part of the name is probably redundant: how about `AddIP` and `AddIPCIDR` as names? Adding addresses themselves together doesn't really make sense, so I don't think there's any potential future ambiguity there.

pkg/net/ip.go
takonomura: My use case is generating network interface configurations, where IP addresses are written in CIDR notation with a prefix length. Since many configuration formats support this form, allowing `AddIPOffset` to take a CIDR string makes it easier to produce such values.

takonomura (commented):
@rogpeppe Thank you for the review. I've made the changes to address your comments.

---

> I just wonder if we'd maybe be better off adding something a bit more general: for example, a way to convert IP / CIDR addresses to bits or a number, and vice versa, and then we'd gain more than just this specific functionality.

I agree that having a more general way to convert IP/CIDR addresses to numbers or bits could be broadly useful.
However, converting IP addresses to numbers can be problematic for IPv6, as IPv6 addresses are typically 39 digits long.
As discussed in https://github.com/cue-lang/cue/issues/3787, this can cause the value to be interpreted as a float during arithmetic operations, which leads to unexpected behavior.

takonomura:
@rogpeppe just a friendly ping! This is ready for re-review when you have a moment.

PR #3822, opened 2025-03-16
internal/encoding/yaml: support alias to scalars in map key (fixes #3821)

### LINKED ISSUE #3821: YAML decoder does not handle aliases in map keys

<pre>
$ cue version
cue version v0.12.0

go version go1.23.7
      -buildmode exe
       -compiler gc
  DefaultGODEBUG asynctimerchan=1,gotypesalias=0,httpservecontentkeepheaders=1,tls3des=1,tlskyber=0,x509keypairleaf=0,x509negativeserial=1
     CGO_ENABLED 1
          GOARCH amd64
            GOOS linux
         GOAMD64 v1
cue.lang.version v0.12.0
</pre>

Yes, and also from main.

### What did you do?
```
cue import my.yaml
```
Where `my.yaml` is:
```
foo: &foo "hello"
*foo : "world"
```

Cue successfully imports the YAML as:
```
foo: "hello"
hello: "world"
```

### What did you see instead?
```
invalid map key: !!str
```

Note that in old versions (e.g 0.8.2) this wouldn't error out, but rather emit this cue file, which is incorrect:
```
foo: "hello"
foo: "world"
```
This is probably a bug in the old YAML decoder (before migrating to yaml.v3)

OmriSteiner:
This makes the decoder follow aliases in YAML map keys. Before this commit the decoder would outright reject these keys as non-scalar keys. This commit adds support specifically for keys which are aliases to scalar values. YAML also permits non-scalar keys such as maps / sequences, but there's no equivalent for those in CUE, so we error out.

fixes #3821

### DIFF
--- a/internal/encoding/yaml/decode.go
+++ b/internal/encoding/yaml/decode.go
@@ -375,9 +375,6 @@ outer:
 			}
 			continue
 		}
-		if yk.Kind != yaml.ScalarNode {
-			return d.posErrorf(yn, "invalid map key: %v", yk.ShortTag())
-		}
 
 		field := &ast.Field{}
 		label, err := d.label(yk)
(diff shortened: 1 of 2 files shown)


mvdan (approved):
Thanks! I'll import into Gerrit shortly.

PR #3860, opened 2025-03-28
fix: Preserve error order when deduplicating errors

### LINKED ISSUE #3859: Cue error deduplication does not preserve order

<pre>
$ cue version
cue version v0.12.0

go version go1.24.1
      -buildmode exe
       -compiler gc
       -trimpath true
  DefaultGODEBUG asynctimerchan=1,gotestjsonbuildtext=1,gotypesalias=0,httpservecontentkeepheaders=1,multipathtcp=0,randseednop=0,rsa1024min=0,tls3des=1,tlsmlkem=0,x509keypairleaf=0,x509negativeserial=1,x509rsacrt=0,x509usepolicies=0
     CGO_ENABLED 1
          GOARCH amd64
            GOOS linux
         GOAMD64 v1
cue.lang.version v0.12.0
</pre>

Yes

### What did you do?

When following the [Schema Definition use case example](https://cuelang.org/docs/concept/schema-definition-use-case/), multiple errors can be returned by `v3.Subsume(v2)`. When printing these errors using `errors.Print`, these errors are displayed in a non-intuitive order instead of inner-to-outer or outer-to-inner because these errors are being sorted as part of a deduplication process.

In the example below I am determining if `#V2` subsumes `#v1` and then printing the errors with `errors.Print`:

```cue
#V1: {
    messages: Foo: crc: 32
}


#V2: {
    messages: Foo: crc: 31 | 33
}
```

the errors displayed are:

```
field Foo not present in {Foo:{crc:32}}:
    ./schema.cue:2:15
field crc not present in {crc:32}:
    ./schema.cue:2:20
field messages not present in {messages:{Foo:{crc:32}}}:
    ./schema.cue:1:6
missing field "messages"
```

A more useful ordering of the errors would be:

```
field crc not present in {crc:32}:
    ./schema.cue:2:20
field Foo not present in {Foo:{crc:32}}:
    ./schema.cue:2:15
field messages not present in {messages:{Foo:{crc:32}}}:
    ./schema.cue:1:6
missing field "messages"
```

which is the order of the underlying array before `errors.RemoveMultiples()` is called.

### What did you see instead?

```
field Foo not present in {Foo:{crc:32}}:
    ./schema.cue:2:15
field crc not present in {crc:32}:
    ./schema.cue:2:20
field messages not present in {messages:{Foo:{crc:32}}}:
    ./schema.cue:1:6
missing field "messages"
```

There is a separate issue where the error messages above are incorrect (the field isn't missing, it's just that `crc: 31 | 33` does not subsume `crc: 32`. I will file a separate issue for that.

GrahamDennis:
Previously when deduplicating errors, the errors were first sorted and
then deduplicated. This lost the ordering of errors which can be useful
for understanding why a value is not compatible with a declaration.

Resolves #3859


mvdan:
Thanks for the patch! Unfortunately this seems to introduce quadratic behavior, which could be problematic as large configurations can run into thousands of errors.

PR #3862, opened 2025-03-29
fix: Improve error message returned by Subsume

### LINKED ISSUE #3861: Error "field X not present in Y" returned by Subsume when field does exist

<pre>
$ cue version
cue version v0.12.0

go version go1.24.1
      -buildmode exe
       -compiler gc
       -trimpath true
  DefaultGODEBUG asynctimerchan=1,gotestjsonbuildtext=1,gotypesalias=0,httpservecontentkeepheaders=1,multipathtcp=0,randseednop=0,rsa1024min=0,tls3des=1,tlsmlkem=0,x509keypairleaf=0,x509negativeserial=1,x509rsacrt=0,x509usepolicies=0
     CGO_ENABLED 1
          GOARCH amd64
            GOOS linux
         GOAMD64 v1
cue.lang.version v0.12.0
</pre>

Yes

### What did you do?

When following the [Schema Definition use case example](https://cuelang.org/docs/concept/schema-definition-use-case/), the error returned can indicate that a field doesn't exist in the target schema instead of correctly indicating that they are not compatible.

In the example below I am determining if `#V2` subsumes `#v1` and then printing the errors with `errors.Print`:

```cue
#V1: {
    messages: Foo: crc: 32
}


#V2: {
    messages: Foo: crc: 31 | 33
}
```

the errors displayed are:

```
field Foo not present in {Foo:{crc:32}}:
    ./schema.cue:2:15
field crc not present in {crc:32}:
    ./schema.cue:2:20
field messages not present in {messages:{Foo:{crc:32}}}:
    ./schema.cue:1:6
missing field "messages"
```

More useful error messages would be

```
31 | 33 does not subsume 32:
    ./schema.cue:2:25
    ./schema.cue:7:25
{Foo:{crc:31 | 33}} does not subsume {Foo:{crc:32}}:
    ./schema.cue:2:15
    ./schema.cue:7:15
{crc:31 | 33} does not subsume {crc:32}:
    ./schema.cue:2:20
    ./schema.cue:7:20
```

### What did you see instead?

```
field Foo not present in {Foo:{crc:32}}:
    ./schema.cue:2:15
field crc not present in {crc:32}:
    ./schema.cue:2:20
field messages not present in {messages:{Foo:{crc:32}}}:
    ./schema.cue:1:6
missing field "messages"
```

GrahamDennis:
Previously when Subsume determined that a field did not subsume a previous field, the error returned was:

```
field x not present in <value>
```

which is simply not true. Now a more useful error is returned in this scenario:

```
31 | 33 does not subsume 32:
    ./schema.cue:2:25
    ./schema.cue:7:25
```

Resolves: #3861

### DIFF
--- a/internal/core/subsume/subsume.go
+++ b/internal/core/subsume/subsume.go
@@ -120,11 +120,9 @@ func (s *subsumer) getError() (err errors.Error) {
 	if s.gt != nil && s.lt != nil {
 		// src := binSrc(token.NoPos, opUnify, s.gt, s.lt)
 		if s.missing != 0 {
-			s.errf("missing field %q", s.missing.SelectorString(c))
-		} else if b, ok := unifyValue(c, s.gt, s.lt).(*adt.Bottom); !ok {
-			s.errf("value not an instance")
+			s.errf("field %q not present in %v", s.missing.SelectorString(c), s.lt)
 		} else {
-			s.errs = errors.Append(s.errs, b.Err)
+			s.errf("%v does not subsume %v", s.gt, s.lt)
 		}
 	}
 	if s.errs == nil {
--- a/internal/core/subsume/testdata/subsume.txtar
+++ b/internal/core/subsume/testdata/subsume.txtar
@@ -3515,93 +3515,118 @@ true
 true
 -- out/TestValues/9/null_⊑__ --
 Errors:
-value not an instance
+null does not subsume _:
+    value:1:4
 
 Result:
 false
 -- out/TestValues/10/int_⊑__ --
 Errors:
-value not an instance
+int does not subsume _:
+    value:1:4
 
 Result:
 false
 -- out/TestValues/11/1_⊑__ --
 Errors:
-value not an instance
+1 does not subsume _:
+    value:1:4
 
 Result:
 false
 -- out/TestValues/12/float_⊑__ --
 Errors:
-value not an instance
+float does not subsume _:
+    value:1:4
 
 Result:
 false
 -- out/TestValues/13/"s"_⊑__ --
 Errors:
-value not an instance
+"s" does not subsume _:
+    value:1:4
 
 Result:
 false
 -- out/TestValues/14/{}_⊑__ --
 Errors:
-value not an instance
+{} does not subsume _:
+    value:1:4
+    value:1:16
 
 Result:
 false
 -- out/TestValues/15/[]_⊑__ --
 Errors:
+[] does not subsume _:
+    value:1:4
+    value:1:16
 list does not subsume _ (type _):
     value:1:16
-value not an instance
 
 Result:
 false
 -- out/TestValues/16/_|__⊑__ --
 Errors:
-value not an instance
+_|_(explicit error (_|_ literal) in source) does not subsume _:
+    value:1:4
+    value:1:18
 
 Result:
 false
 -- out/TestValues/17/_|__⊑_null --
 Errors:
-value not an instance
+_|_(explicit error (_|_ literal) in source) does not subsume null:
+    value:1:4
+    value:1:12
 
 Result:
 false
 -- out/TestValues/18/_|__⊑_int --
 Errors:
-value not an instance
+_|_(explicit error (_|_ literal) in source) does not subsume int:
+    value:1:4
+    value:1:12
 
 Result:
 false
 -- out/TestValues/19/_|__⊑_1 --
 Errors:
-value not an instance
+_|_(explicit error (_|_ literal) in source) does not subsume 1:
+    value:1:4
+    value:1:12
 
 Result:
 false
 -- out/TestValues/20/_|__⊑_float --
 Errors:
-value not an instance
+_|_(explicit error (_|_ literal) in source) does not subsume float:
+    value:1:4
+    value:1:12
 
 Result:
 false
 -- out/TestValues/21/_|__⊑_"s" --
 Errors:
-value not an instance
+_|_(explicit error (_|_ literal) in source) does not subsume "s":
+    value:1:4
+    value:1:12
 
 Result:
 false
 -- out/TestValues/22/_|__⊑_{} --
 Errors:
-value not an instance
+_|_(explicit error (_|_ literal) in source) does not subsume {}:
+    value:1:4
+    value:1:12
 
 Result:
 false
 -- out/TestValues/23/_|__⊑_[] --
 Errors:
-value not an instance
+_|_(explicit error (_|_ literal) in source) does not subsume []:
+    value:1:4
+    value:1:12
 
 Result:
 false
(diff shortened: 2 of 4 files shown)


mvdan:
Thanks for the patch. Could you add a test? Interestingly, we don't seem to have no tests for "field ... not present in ...". I would imagine this test would go somewhere under `cue/testdata`. Best if you add the test in a first commit, then the second commit can update the expected output and show the improvement in behavior.

GrahamDennis:
> Thanks for the patch. Could you add a test? Interestingly, we don't seem to have no tests for "field ... not present in ...". I would imagine this test would go somewhere under `cue/testdata`. Best if you add the test in a first commit, then the second commit can update the expected output and show the improvement in behavior.

Thanks for the feedback. I've done this now.

The first commit contains just the new tests and contains [some examples that I think have incorrect error messages](https://github.com/cue-lang/cue/pull/3862/commits/18562818bebee5daaca3d0ff535d1b79647ad383#diff-6924da255b9a20aac8c5ae301ea8c2fabcbdd55f2472fef4e8cac353d48ed59dR4204-R4243):

```
-- out/TestValues/407/{foo:_1}_⊑_{foo?:_1} --
Errors:
field foo not present in {foo?:1}:
    value:1:17
missing field "foo"

Result:
false
```

After the second commit, the updated error message is:

```
-- out/TestValues/407/{foo:_1}_⊑_{foo?:_1} --
Errors:
1 does not subsume 1:
    value:1:10
    value:1:24
1 does not subsume 1?:
    value:1:10
    value:1:24

Result:
false
```

edit: I fixed a bug in my change with the error messages generated by the v3 evaluator vs the v2 evaluator.

internal/core/subsume/vertex.go
```diff
@@ -170,11 +170,19 @@ func (s *subsumer) vertices(x, y *adt.Vertex) bool {
 			continue
 		}
 
-		s.missing = f
-		s.gt = a
-		s.lt = y
+		if b == nil {
```
mpvl, line 173: I'm unclear how this condition can ever be true: if `b == nil`, b will be set to a non-nil value in the block above.

internal/core/subsume/vertex.go
```diff
@@ -373,11 +381,18 @@ func (s *subsumer) verticesDev(x, y *adt.Vertex) bool {
...
```
mpvl: ditto

internal/core/subsume/vertex.go
mpvl: I'm unclear how this condition can ever be true: if b == nil, b will be set to a non-nil value in the block above.

internal/core/subsume/testdata/subsume.txtar
```diff
@@ -0,0 +1,4837 @@
+-- out/TestValues/0/__⊑__/v2 --
```
mpvl: I don't see any test inputs here. Am I missing something?

internal/core/subsume/vertex.go
```diff
@@ -170,11 +173,11 @@ func (s *subsumer) vertices(x, y *adt.Vertex) bool {
 			continue
 		}
 
-		s.missing = f
 		s.gt = a
-		s.lt = y
+		s.lt = b
+
+		s.errf("%v%s does not subsume %v%s", a, a.ArcType.Suffix(), b, b.ArcType.Suffix())
```
GrahamDennis, line 179: This has been modified to improve the error for situations like: `{foo:_1}_⊑_{foo?:_1}`. Currently the error is:

```
1 does not subsume 1
```

But there will now be a new error of

```
1 does not subsume 1?
```

This may be considered an abuse of syntax, but it's a starting point for discussion.

internal/core/subsume/testdata/subsume.txtar
GrahamDennis: This is driven by the existing test cases in https://github.com/cue-lang/cue/pull/3862/files#diff-5c570d819733910b5664134701d7cb95c9efbf2261b72b32b3ff5339ef18bb27L52-R77.
What I wanted to do was re-use all of those test cases and then use the txtar tooling for rendering the error messages.

internal/core/subsume/vertex.go
GrahamDennis: You are correct. I have removed this and also set the `s.missing` et al fields inside the block above.

internal/core/subsume/vertex.go
GrahamDennis: You are correct. I have removed this and also set the `s.missing` et al fields inside the block above.

GrahamDennis:
Ping on this PR?

mvdan:
We haven't forgotten about your PRs; we're just in a bit of a crunch for the v0.13 release with the new evaluator :) Please bear with us, and thank you for your patience!

GrahamDennis:
ping @mvdan ? Any update here now that 0.13 is out?

GrahamDennis:
ping @mvdan / @mpvl ?

mvdan:
Thanks; we haven't forgotten about this. Please understand that we have a number of competing priorities at the moment, and a small team.

internal/core/subsume/testdata/subsume.txtar
mpvl: We typically want to keep the tests together. So either put input an output in txtar, or do everything in a table.

Also, we prefer not to change the general design of test. It is a lot of code to review! Finally, we want to separate out semantic from structural changes. So if we want to change the generic test structure, this should be a separate CL.

internal/core/subsume/subsume.go
```diff
@@ -120,11 +120,9 @@ func (s *subsumer) getError() (err errors.Error) {
 	if s.gt != nil && s.lt != nil {
 		// src := binSrc(token.NoPos, opUnify, s.gt, s.lt)
 		if s.missing != 0 {
-			s.errf("missing field %q", s.missing.SelectorString(c))
-		} else if b, ok := unifyValue(c, s.gt, s.lt).(*adt.Bottom); !ok {
-			s.errf("value not an instance")
+			s.errf("field %q not present in %v", s.missing.SelectorString(c), s.lt)
```
mpvl, line 123: This commit contains rephrasing of error messages, as well as logic changes. This makes it non-trivial to see which tests change because of what. These should be separated out in different CLs.

mpvl (commented):
Thanks for your updates and pardon the delay.

Firstly, this is quite a complex change. For anything but trivial changes we typically require moving over to Gerrit, for several reasons:
1) all reviews documented in a single place, 
2) code review capabilities of GH are rather poor (for instance no fine-grained colored diffing), which makes it really hard to review,
3) allow commenting on commit messages (for instance, headers do not follow our pattern).

I made some comments, but I suggest moving over to Gerrit henceforth.

mpvl:
I've updated the testing setup for TestValues. It now tests for the particular error message. You can run 
```
CUE_UPDATE=1 go test ./internal/core/subsume  
```
To update the generated error messages.

The reviews for this are still pending, though: https://review.gerrithub.io/c/cue-lang/cue/+/1219851/1

mvdan:
Friendly nudge @GrahamDennis, not sure if you saw Marcel's replies above. You can see https://github.com/cue-lang/cue/blob/master/CONTRIBUTING.md in terms of sending patches on Gerrit.

mvdan:
I'll close this PR for now, given that it's inactive; please feel free to open a new PR with Marcel's feedback above :)

PR #3863, opened 2025-03-30
fix call function with a nil value error

alingse:
see also https://go-review.googlesource.com/c/tools/+/661695


mvdan:
Thanks for the patch! Please note that this code is copied/vendored from https://cs.opensource.google/go/x/tools, so we didn't write the code nor do we really maintain it right now. You should direct your fix there, if the bug is still present upstream.

I'll leave it to @myitcv to decide if the fix here is useful, i.e. if this bit of code is used by `cue lsp` or not.

cuematthew:
Hi @alingse . Thank you for the PR. I'd like to merge this, but we need you to follow the instructions for providing a [Developer Certificate of Origin](https://github.com/cue-lang/cue/blob/master/CONTRIBUTING.md#asserting-a-developer-certificate-of-origin) before we can proceed.

alingse:
@cuematthew done, I push a signed commit

PR #3987, opened 2025-07-07
internal/encoding/yaml: encode YAML anchors as CUE definitions

### LINKED ISSUE #3818: encoding/yaml: decode anchors without expanding them
YAML has anchors, allowing to name nodes and reference them later in the document without repeating their content:

```
$ cat f.yaml
foo: &name "long expression"
bar: *name
```

Currently, our YAML decoder fully expands these references, even when importing as CUE:

```
$ cue def f.yaml
foo: "long expression"
bar: "long expression"
$ cue import f.yaml
$ cat f.cue
foo: "long expression"
bar: "long expression"
```

We could be a bit smarter and translate some of these anchors to CUE references, such as:

```
$ cat f.cue
name=foo: "long expression"
bar: name
$ cue export f.cue
{
    "foo": "long expression",
    "bar": "long expression"
}
```

However, note that anchors won't always translate one-to-one to CUE references. One example is scoping; YAML anchors always work as long as the anchor is defined earlier in the document, whereas in CUE, references must follow scoping rules, meaning that an earlier alias or field might no longer be in scope.

```
$ cat f.yaml
foo:
  nested: &name "long expression"
bar:
  nested: *name
$ cue export f.yaml
{
    "foo": {
        "nested": "long expression"
    },
    "bar": {
        "nested": "long expression"
    }
}
$ cat f.cue
foo: name=nested: "long expression"
bar: nested: name
$ cue export f.cue
foo: unreferenced alias or let clause name:
    ./f.cue:1:6
bar.nested: reference "name" not found:
    ./f.cue:2:14
```

One potential solution would be to always place anchors at the top level as let fields or definitions, such as the following, which does work, but somewhat loses the semblance with YAML:

```
let name = "long expression"
foo: nested: name
bar: nested: name
```

OmriSteiner:
Fixes #3818

There's one small issue open to handle. Tests are currently failing with:

```
--- FAIL: TestUnmarshalErrors (0.00s)
    --- FAIL: TestUnmarshalErrors/test_7:_"a:_&a\n__b:_*a\n" (0.00s)
        decode_test.go:971:   yaml:
            a: &a
              b: *a
        decode_test.go:977: got <nil>; want test.yaml:2: anchor "a" value contains itself; (value #a: {
                b: #a
            }
            a: #a)
--- FAIL: TestDecoderErrors (0.00s)
    --- FAIL: TestDecoderErrors/test_7:_"a:_&a\n__b:_*a\n" (0.00s)
        decode_test.go:986:   yaml:
            a: &a
              b: *a
        decode_test.go:988: got <nil>; want test.yaml:2: anchor "a" value contains itself
```

I could change the code to match the old behavior of rejecting such YAML documents, since the resulting CUE document is invalid.
However, I think the original reason we error out for such document was that it would lead to endless recursion in the parser. This is no longer the case, so I'm debating whether the correct fix would be to error out, or change the test.

### DIFF
--- a/internal/encoding/yaml/decode.go
+++ b/internal/encoding/yaml/decode.go
@@ -64,6 +64,15 @@ type decoder struct {
 
 	// forceNewline ensures that the next position will be on a new line.
 	forceNewline bool
+
+	// anchorFields contains the anchors that are gathered as we walk the YAML nodes.
+	// these are only added to the AST when we're done processing the whole document.
+	anchorFields []ast.Field
+	// anchorNames map anchor nodes to their names.
+	anchorNames map[*yaml.Node]string
+	// anchorTakenNames keeps track of anchor names that have been taken.
+	// It is used to ensure unique anchor names.
+	anchorTakenNames map[string]struct{}
 }
 
 // TODO(mvdan): this can be io.Reader really, except that token.Pos is offset-based,
@@ -83,9 +92,11 @@ func NewDecoder(filename string, b []byte) *decoder {
 	tokFile := token.NewFile(filename, 0, len(b)+1)
 	tokFile.SetLinesForContent(b)
 	return &decoder{
-		tokFile:     tokFile,
-		tokLines:    append(tokFile.Lines(), len(b)),
-		yamlDecoder: *yaml.NewDecoder(bytes.NewReader(b)),
+		tokFile:          tokFile,
+		tokLines:         append(tokFile.Lines(), len(b)),
+		yamlDecoder:      *yaml.NewDecoder(bytes.NewReader(b)),
+		anchorNames:      make(map[*yaml.Node]string),
+		anchorTakenNames: make(map[string]struct{}),
 	}
 }
 
@@ -176,24 +187,35 @@ func Unmarshal(filename string, data []byte) (ast.Expr, error) {
 	return n, nil
 }
 
-func (d *decoder) extract(yn *yaml.Node) (ast.Expr, error) {
-	d.addHeadCommentsToPending(yn)
-	var expr ast.Expr
-	var err error
+func (d *decoder) extractNoAnchor(yn *yaml.Node) (ast.Expr, error) {
 	switch yn.Kind {
 	case yaml.DocumentNode:
-		expr, err = d.document(yn)
+		return d.document(yn)
 	case yaml.SequenceNode:
-		expr, err = d.sequence(yn)
+		return d.sequence(yn)
 	case yaml.MappingNode:
-		expr, err = d.mapping(yn)
+		return d.mapping(yn)
 	case yaml.ScalarNode:
-		expr, err = d.scalar(yn)
+		return d.scalar(yn)
 	case yaml.AliasNode:
-		expr, err = d.alias(yn)
+		return d.referenceAlias(yn)
 	default:
 		return nil, d.posErrorf(yn, "unknown yaml node kind: %d", yn.Kind)
 	}
+}
+
+func (d *decoder) extract(yn *yaml.Node) (ast.Expr, error) {
+	d.addHeadCommentsToPending(yn)
+
+	var expr ast.Expr
+	var err error
+
+	if yn.Anchor == "" {
+		expr, err = d.extractNoAnchor(yn)
+	} else {
+		expr, err = d.anchor(yn)
+	}
+
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


mvdan:
If recursive anchors like that no longer fail, what CUE do they result in?

OmriSteiner:
> If recursive anchors like that no longer fail, what CUE do they result in?

You can see in the error above:
```
#a: {
    b: #a
}
a: #a
```

Still recursive, so it's invalid. But the encoder doesn't recurse infinitely.

mvdan:
Ah I see. That's fine; I don't imagine these cases will actually show up in practice often. Indeed I think we avoided cycles because of endless recursion.

OmriSteiner:
On the one hand it makes no sense to emit an invalid CUE file.
But on the other hand - perhaps there's a use-case I'm missing here?

OmriSteiner:
> Ah I see. That's fine; I don't imagine these cases will actually show up in practice often. Indeed I think we avoided cycles because of endless recursion.

Fine, then I'll edit the test.

internal/encoding/yaml/decode.go
```diff
@@ -649,11 +705,58 @@ func (d *decoder) alias(yn *yaml.Node) (ast.Expr, error) {
...
 	d.extractingAliases[yn] = true
 	var node ast.Expr
-	node, err := d.extract(yn.Alias)
+	node, err := d.extract_no_anchor(yn.Alias)
 	delete(d.extractingAliases, yn)
 	return node, err
 }
 
+// Replace an alias with a reference to the identifier of its anchor.
+func (d *decoder) reference_alias(yn *yaml.Node) (ast.Expr, error) {
```
mvdan: in Go we typically use camelCase, not snake_case.

internal/encoding/yaml/decode.go
```diff
@@ -639,7 +692,10 @@ func (d *decoder) makeNum(yn *yaml.Node, val string, kind token.Token) (expr ast
 	return expr
 }
 
-func (d *decoder) alias(yn *yaml.Node) (ast.Expr, error) {
+// Expands an alias node in place, returning the expanded node.
```
mvdan: Please skim https://go.dev/blog/godoc - when documenting fields or methods, or really any sort of declaration, you should start by naming what it is you're documenting. Like: `// inlineAlias expands an alias...`

internal/encoding/yaml/decode.go
```diff
@@ -324,7 +345,39 @@ func (d *decoder) document(yn *yaml.Node) (ast.Expr, error) {
...
 		return nil, d.posErrorf(yn, "yaml document nodes are meant to have one content node but have %d", n)
 	}
-	return d.extract(yn.Content[0])
+
+	expr, err := d.extract(yn.Content[0])
+	if err != nil {
+		return nil, err
+	}
+
+	return d.addAnchorNodes(expr)
```
mvdan, line 355: rather than buffering these for the entire document and adding them at the very top, it would be better to add them closer to where they are used. if an anchor is declared in the middle of the document, we should place the definition in roughly the same spot, and not at the very start or end of the whole file.

OmriSteiner: I think we discussed this when we create this issue several months ago.

Since CUE definitions are scoped, we can't place the anchor definition exactly at the scope of the original anchor, since there may be aliases outside this scope.

So there are two options to go about this, the simpler would be to lift all anchor definitions to the top-level of the CUE document. But then - is it any different than putting everything at the top / bottom?
The more complex solution would be to find the inner-most scope containing all anchor aliases. But since this code is recursive, doing so without buffering would be quite complex - is it worth it?

mvdan: To be clear, I don't mean to place them in an inner scope - the top level file scope is fine. I just mean placing them relatively close to their original position - in my example above, near the middle of the file - rather than at the start or end of the whole file.

OmriSteiner:
I pushed some changes to make the tests pass.
I also found an additional edge case I'll have to fix, so currently working on that.

internal/encoding/yaml/decode.go
OmriSteiner: I have an implementation for this, but it's not really nice.
I can push this as a separate commit if you want to have a look.
Basically you have to add a recursion parameter to differentiate between when you're handling a top-level map as opposed to any other map.
IMO this just makes the code more complicated for little actual benefit.

OmriSteiner:
> I pushed some changes to make the tests pass. I also found an additional edge case I'll have to fix, so currently working on that.

Alright, that should be sorted. Had a positioning issue where I accidentally called `d.pos` for the anchor reference *after* extracting the node itself.

internal/encoding/yaml/decode.go
OmriSteiner: Anyway, I pushed the patch. But I think it's unwise to merge it like this.
If you have any suggestions for a cleaner implementation I'm up for it.
But in the current form, I think keeping things simple is more important than being marginally better.

mvdan:
@OmriSteiner the tests are failing, not sure if you saw that?

OmriSteiner:
> @OmriSteiner the tests are failing, not sure if you saw that?

Yes, I did.
As I wrote:

> Anyway, I pushed the patch. But I think it's unwise to merge it like this.
> If you have any suggestions for a cleaner implementation I'm up for it.
> But in the current form, I think keeping things simple is more important than being marginally better.

I only pushed the additional commit for you to look at. If you think this is worth the tradeoff with implementation complexity, then I can fix the tests. Otherwise I think it's better if I remove the last commit here (and then the tests pass).

mvdan:
Ah, I had misunderstood. And apologies for the slowness. Yes, I think your latest commit is a perfectly fine approach. If you can fix the tests, I'll do a final review :)

mvdan (changes requested):
Marking as "request changes" as a reminder to myself that this needs CI to be fixed before I review again.

mvdan (approved):
Thanks!

mvdan:
importing into Gerrit here: https://cue.gerrithub.io/c/cue-lang/cue/+/1223900

I tweaked the commit message a bit as well, assuming that's okay with you.

mvdan:
@rogpeppe correctly spotted that the use of definitions breaks valid use cases today, so I need to take a closer look. See the Gerrit patch discussion.

mvdan:
There was a bit of confusion on the Gerrit side; this patch is now at https://cue.gerrithub.io/c/cue-lang/cue/+/1224165.

OmriSteiner:
> There was a bit of confusion on the Gerrit side; this patch is now at https://cue.gerrithub.io/c/cue-lang/cue/+/1224165.

That is correct. Let me know if there is anything required from my end.
I no longer have the need for this feature, but I'm willing to make any changes required to merge this.
Alternatively we could also just close this PR, but I think this may be a useful feature for some.

mvdan:
@OmriSteiner apologies that we didn't move faster here. I still plan to address issue #3818, and I'll credit you for your part in it.

PR #4003, opened 2025-07-25
yaml/validate: Improve error reporting

### LINKED ISSUE #4002: yaml.Validate missing suberrors

<pre>
0.13.0
</pre>

Yes

### What did you do?

Consider the following code example and observe the stdout.
```go
package main

import (
	"fmt"

	"cuelang.org/go/cue/errors"
	"cuelang.org/go/encoding/yaml"
)

const cueSource = `foo: 1 | 2`

func main() {
	ctx := cuecontext.New()
	schema := ctx.CompileString(cueSource)

	data := "foo: 3"
	if err := yaml.Validate([]byte(data), schema); err != nil {
		fmt.Println("yaml.Validate(..)")
		fmt.Println(errors.Details(err, nil))
	}

	yamlFile, _ := yaml.Extract("data.yml", data)
	yamlAsCUE := ctx.BuildFile(yamlFile)
	unified := schema.Unify(yamlAsCUE)
	if err := unified.Validate(); err != nil {
		fmt.Println("schema.Unify(..).Validate()")
		fmt.Println(errors.Details(err, nil))
	}
}
```

Same error for `yaml.Validate` and `schema.Unify(yaml).Validate()`:
```
yaml.Validate(..)
foo: 2 errors in empty disjunction:
foo: conflicting values 1 and 3:
    1:6
    data.yml:1:6
foo: conflicting values 2 and 3:
    1:10
    data.yml:1:6

schema.Unify(..).Validate()
foo: 2 errors in empty disjunction:
foo: conflicting values 1 and 3:
    1:6
    data.yml:1:6
foo: conflicting values 2 and 3:
    1:10
    data.yml:1:6
```

### What did you see instead?
`yaml.Validate(..)` is missing error details
```
yaml.Validate(..)
foo: 2 errors in empty disjunction:

schema.Unify(..).Validate()
foo: 2 errors in empty disjunction:
foo: conflicting values 1 and 3:
    1:6
    data.yml:1:6
foo: conflicting values 2 and 3:
    1:10
    data.yml:1:6
```

_I think I've found the issue, will submit PR shortly_

kilpkonn:
This makes `yaml.Validate(data, schema)` work more similarly to what one would likely do manually (`schema.Unify(data).Validate()`).

Namely this improves error reporting.

Closes https://github.com/cue-lang/cue/issues/4002

### DIFF
--- a/cue/testdata/cycle/builtins.txtar
+++ b/cue/testdata/cycle/builtins.txtar
@@ -796,24 +796,24 @@ Result:
       }
     }
     yamlVal: (_|_){
-      // [eval]
+      // [structural cycle]
       t1: (_|_){
-        // [eval]
+        // [structural cycle]
         x: (_|_){
-          // [eval]
+          // [structural cycle]
           y: (_|_){
-            // [eval] selfCycle.yamlVal.t1.x.y: invalid value "{}" (does not satisfy encoding/yaml.Validate): error in call to encoding/yaml.Validate: structural cycle:
+            // [structural cycle] selfCycle.yamlVal.t1.x.y: invalid value "{}" (does not satisfy encoding/yaml.Validate): error in call to encoding/yaml.Validate: structural cycle:
             //     ./yamlcycle.cue:24:8
             //     ./yamlcycle.cue:25:8
           }
         }
       }
       t2: (_|_){
-        // [eval]
+        // [structural cycle]
         x: (_|_){
-          // [eval]
+          // [structural cycle]
           y: (_|_){
-            // [eval] selfCycle.yamlVal.t2.x.y: invalid value "{}" (does not satisfy encoding/yaml.Validate): error in call to encoding/yaml.Validate: structural cycle:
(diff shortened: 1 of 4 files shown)


mvdan:
Thanks! This change aligns with `pkg/encoding/json.Validate` so I think it's correct.

Please add a test; see `pkg/encoding/yaml/testdata/validate.txtar`.

Ideally we add the same sort of test to `pkg/encoding/json/testdata/gen.txtar` under the "validate" section, to sanity check that the two are consistent.

Typically, when adding a regression test, we add it before the fix to record the "before" output, via `CUE_UPDATE=1 go test ./...`, and then the second commit applies the fix and improves the expected output in the test. That way, it's very obvious how the fix is helping.

kilpkonn:
Should be good now

mvdan (approved):
Thanks! Will import to Gerrit and merge.

mvdan:
Imported as https://review.gerrithub.io/c/cue-lang/cue/+/1219285; cc @myitcv for review

PR #4027, opened 2025-08-13
encoding/toml: fix decoding subtable duplicate keys

### LINKED ISSUE #3609: encoding/toml:  incorrect `duplicate key` error for tables nested within an array of tables

<pre>
$ cue version
cue version v0.11.0

go version go1.23.3
      -buildmode exe
       -compiler gc
       -trimpath true
     CGO_ENABLED 0
          GOARCH amd64
            GOOS linux
         GOAMD64 v1
cue.lang.version v0.11.0
</pre>

Yes.

### What did you do?
```txtar
# eval cue example
exec cue eval example.cue

# eval equivalent toml example
exec cue eval example.toml
-- example.cue --
blocks: [
	{
		x: 42
		properties: {
			style: "plain"
		}
	},
	{
		x: 100
		properties: {
			style: "special"
		}
	},
]
-- example.toml --
[[blocks]]
x = 42

[blocks.properties]
style = 'plain'

[[blocks]]
x = 100

[blocks.properties]
style = 'special'
```

Both `cue eval` commands run successfully with no errors.

### What did you see instead?

```
$ cue eval example.toml
duplicate key: blocks.properties:
    ./example.toml:10:2
```

This is incorrect according to the [TOML spec (v1.0.0)](https://toml.io/en/v1.0.0#array-of-tables), which states:

> Any reference to an array of tables points to the most recently defined table element of the array. This allows you to define sub-tables, and even sub-arrays of tables, inside the most recent table.

This issue has the undesirable effect of breaking, for example, the `cue eval` and `cue vet` commands for TOML files containing such data.

theoretick:
Previously, subtables with duplicate keys fail to be decoded even though they represent valid TOML. Instead we should track seenTableKeys by the array index to ensure duplication is local to the given object

Fixes https://github.com/cue-lang/cue/issues/3609

### DIFF
--- a/encoding/toml/decode.go
+++ b/encoding/toml/decode.go
@@ -202,11 +202,28 @@ func (d *Decoder) nextRootNode(tnode *toml.Node) error {
 	case toml.Table:
 		// Tables always begin a new line.
 		key, keyElems := d.decodeKey("", tnode.Key())
+
+		// Check if this table is a subtable of an existing array element
+		array := d.findArrayPrefix(key)
+		var actualKey string
+		if array != nil { // [last_array.new_table]
+			if array.rkey == key {
+				return d.nodeErrf(tnode.Child(), "cannot redeclare table array %q as a table", key)
+			}
+			// For subtables within array elements, we need to use the current array element's key
+			// to avoid false duplicate key errors between different array elements
+			subKey := key[len(array.rkey)+1:] // Remove the array prefix and dot
+			actualKey = fmt.Sprintf("%s.%d.%s", array.rkey, len(array.list.Elts)-1, subKey)
+
+		} else {
+			actualKey = key
+		}
+
 		// All table keys must be unique, including for the top-level table.
-		if d.seenTableKeys[key] {
+		if d.seenTableKeys[actualKey] {
 			return d.nodeErrf(tnode.Child(), "duplicate key: %s", key)
 		}
-		d.seenTableKeys[key] = true
+		d.seenTableKeys[actualKey] = true
 
 		// We want a multi-line struct with curly braces,
 		// just like TOML's tables are on multiple lines.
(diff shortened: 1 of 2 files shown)


encoding/toml/decode.go
```diff
@@ -202,11 +202,28 @@ func (d *Decoder) nextRootNode(tnode *toml.Node) error {
...
+		var actualKey string
+		if array != nil { // [last_array.new_table]
+			if array.rkey == key {
+				return d.nodeErrf(tnode.Child(), "cannot redeclare table array %q as a table", key)
+			}
+			// For subtables within array elements, we need to use the current array element's key
+			// to avoid false duplicate key errors between different array elements
+			arrayElementKey := array.rkey + "." + strconv.Itoa(len(array.list.Elts)-1)
+			subKey := key[len(array.rkey)+1:] // Remove the array prefix and dot
+			actualKey = arrayElementKey + "." + subKey
```
mvdan: do all the string concatenation in one go, e.g. `array.rkey + "." + strconv.Itoa(len(array.list.Elts)-1) + "." + subKey`, as two separate string concatenations will result in extra allocations.

or if you ask me, I'd personally use Sprintf, like:

    fmt.Sprintf("%s.%d.%s", array.rkey, len(array.ListElts)-1, subKey)

mvdan (approved):
Thanks! This looks fine - just a minor nit about the string concatenation.

encoding/toml/decode.go
theoretick: I'm good with Sprintf. Done with c7f580fe

theoretick:
@mvdan I'm not sure what's causing this test failure 🤔 would you mind taking a look?

mvdan:
@theoretick my best guess is that you need to rebase on master. I don't see how your changes could cause such a failure other than not being up to date with master.

mvdan:
I'll rebase as I import the patch into Gerrit, assuming that will fix CI.

mvdan:
Imported as https://review.gerrithub.io/c/cue-lang/cue/+/1221525.

PR #4034, opened 2025-08-19
encoding/protobuf/parse: fix crashing on builtin package imports

### LINKED ISSUE #4033: Protobuf builtin packages do not work properly

<pre>
cue version v0.14.1

go version go1.24.6
      -buildmode exe
       -compiler gc
       -trimpath true
     CGO_ENABLED 0
          GOARCH arm64
            GOOS darwin
         GOARM64 v8.0
cue.lang.version v0.14.1

</pre>

Yes, this is the most recent version released

### What did you do?

I have the following proto definition in a file named `builtinpkg.proto`:

```protobuf
syntax = "proto3";

package builtinpkg;

import "google/protobuf/timestamp.proto";

message DefaultPkgTestMsg {
  google.protobuf.Timestamp observed_at = 1;
}
```

When I run `cue import proto builtinpkg.proto`, I get an error

I expected that this should work without providing an import path for `google/protobuf/timestamp.proto` because this is a builtin package and should output a valid cue file

### What did you see instead?

```
could not find import "google/protobuf/timestamp.proto":
    ./builtinpkg.proto:5:1
```

hiporox:
This commit fixes an issue where importing

"google/protobuf/timestamp.proto" or other builtin packages causes

the go proto extractor to crash.

By checking if the import is a builtin package before failing.

Additionally, a test case is added to ensure that this behavior works

Resolves https://github.com/cue-lang/cue/issues/4033

### DIFF
--- a/encoding/protobuf/parse.go
+++ b/encoding/protobuf/parse.go
@@ -313,14 +313,14 @@ func (p *protoConverter) doImport(v *proto.Import) error {
 		break
 	}
 
+	if !p.mapBuiltinPackage(v.Position, v.Filename, filename == "") {
+		return nil
+	}
+
 	if filename == "" {
 		err := errors.Newf(p.toCUEPos(v.Position), "could not find import %q", v.Filename)
 		p.state.addErr(err)
 		return err
-	}
-
-	if !p.mapBuiltinPackage(v.Position, v.Filename, filename == "") {
-		return nil
 	}
 
 	imp, err := p.state.parse(filename, nil)
--- a/encoding/protobuf/protobuf_test.go
+++ b/encoding/protobuf/protobuf_test.go
@@ -37,6 +37,7 @@ func TestExtractDefinitions(t *testing.T) {
 		"mixer/v1/attributes.proto",
 		"mixer/v1/config/client/client_config.proto",
 		"other/trailcomment.proto",
+		"other/builtinpkg.proto",
 	}
 	for _, file := range testCases {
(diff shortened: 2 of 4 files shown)


mvdan:
I was a bit confused by this patch, because adding the test without the fix did not result in a test failure. I can see now why - `encoding/protobuf/testdata` is added to the search path for protobuf imports, and that contains files such as `google/protobuf/timestamp.proto`.

These imports still get mapped to builtin types when the file exists on disk - the code just required the file to exist on disk. I assume that was your original issue - you probably do not have `google/protobuf/timestamp.proto` on disk. And I don't think we should require that - we are not using such a file from disk anyway, as we are translating it into builtin types.

Hence I think your patch is correct, but it only really makes sense if we delete these unnecessary testdata files too, as otherwise the test doesn't actually act as a proper regression test.

I've also figured out that we had two ways to error with "import not found", meaning that one was entirely useless. I'm cleaning up the code accordingly as well.

I'll import your patch to Gerrit as-is, and then I'll do a second patchset with my suggested changes.

mvdan (approved):

mvdan:
I also realise now that a new test is unnecessary - we already have other tests importing `timestamp.proto`.

mvdan:
Imported as https://review.gerrithub.io/c/cue-lang/cue/+/1221533. You can see my changes on top of yours by diffing between patchsets 1 and 2: https://review.gerrithub.io/c/cue-lang/cue/+/1221533/1..2

I added myself as a co-author accordingly.

PR #4036, opened 2025-08-19
encoding/protobuf/parse: fix failing to find explicit names in same package

### LINKED ISSUE #4035: Protobufs with fully qualified in package names fail

<pre>
$ cue version

cue version v0.14.1

go version go1.24.6
      -buildmode exe
       -compiler gc
       -trimpath true
     CGO_ENABLED 0
          GOARCH arm64
            GOOS darwin
         GOARM64 v8.0
cue.lang.version v0.14.1
</pre>

Yes, this is the latest release

### What did you do?

I have the following proto file (`explicit_msg.proto`):
```
syntax = "proto3";

package istio.io.api.other.explicit_names;

message ExplicitMsg {
  istio.io.api.other.explicit_names.ExplicitNestedMsg nested_msg = 1;
}

message ExplicitNestedMsg {
  string value = 1;
}
```

If I used `cue import proto explicit_msg.proto`, I get an error

I can do the same thing an enum instead of a nested message to same result

For this command to work and output a valid cue file

### What did you see instead?

```
error parsing protobuf: name "istio.io.api.other.explicit_names.ExplicitNestedMsg" not found:
    ./explicit_names.proto:6:3
```

hiporox:
This commit makes it so that fully qualified names

that are defined in the same file

are resolved properly.

It also adds tests to confirm that this behavior works

Resolves: https://github.com/cue-lang/cue/issues/4035

### DIFF
--- a/encoding/protobuf/parse.go
+++ b/encoding/protobuf/parse.go
@@ -277,7 +277,11 @@ func (p *protoConverter) resolveTopScope(pos scanner.Position, name string, opti
 		if k == -1 {
 			i = len(name)
 		}
-		if m, ok := p.scope[0][name[:i]]; ok {
+		curName := name[:i]
+		if local, ok := strings.CutPrefix(curName, p.protoPkg+"."); ok {
+			curName = local
+		}
+		if m, ok := p.scope[0][curName]; ok {
 			if m.pkg != nil {
 				p.imported[m.pkg.qualifiedImportPath()] = true
 			}
@@ -292,6 +296,7 @@ func (p *protoConverter) resolveTopScope(pos scanner.Position, name string, opti
 			ast.SetPos(expr, p.toCUEPos(pos))
 			return expr
 		}
+
 	}
 	failf(pos, "name %q not found", name)
 	return nil
--- a/encoding/protobuf/protobuf_test.go
+++ b/encoding/protobuf/protobuf_test.go
@@ -37,6 +37,7 @@ func TestExtractDefinitions(t *testing.T) {
 		"mixer/v1/attributes.proto",
 		"mixer/v1/config/client/client_config.proto",
 		"other/trailcomment.proto",
+		"other/full_references.proto",
 	}
 	for _, file := range testCases {
 		t.Run(file, func(t *testing.T) {
--- a/encoding/protobuf/testdata/full_references.proto.out.cue
+++ b/encoding/protobuf/testdata/full_references.proto.out.cue
@@ -1,0 +1,13 @@
+package full_references
+
+#FullReferenceMsg: {
+	nestedMsg?: #FullReferenceNestedMsg @protobuf(1,istio.io.api.other.full_references.FullReferenceNestedMsg,name=nested_msg)
+}
+
+#FullReferenceNestedMsg: {
+
+	#FullReferenceDoubleNestedMsg: {
+		value?: string @protobuf(1,string)
+	}
+	nestedMsg?: #FullReferenceNestedMsg.#FullReferenceDoubleNestedMsg @protobuf(1,istio.io.api.other.full_references.FullReferenceNestedMsg.FullReferenceDoubleNestedMsg,name=nested_msg)
+}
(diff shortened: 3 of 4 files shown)


encoding/protobuf/parse.go
```diff
@@ -245,6 +245,9 @@ func (p *protoConverter) addNames(elems []proto.Visitee) {
 		}
 		sym := strings.Join(append(p.path, name), ".")
 		p.symbols[sym] = true
+		qualifiedSym := strings.Join([]string{p.protoPkg, name}, ".")
+		p.symbols[qualifiedSym] = true
+		p.addRef(pos, qualifiedSym, func() ast.Expr { return ast.NewIdent("#" + name) })
```
mvdan: AFAICT you only need to call `addRef`, adding to `symbols` is unnecessary. and potentially harmful, as it is about what symbols the package exports.

I also think adding directly to the scope via `addRef` is a bit weird. Rather than duplicating every name in the scope, what I would do is modify `resolveTopScope` such that when it is looking for `fooBar`, it searches for both `fooBar` (the short name) and `current.package.name.fooBar` (the full name).

encoding/protobuf/testdata/istio.io/api/other/explicit_names.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
```
mvdan: can you name this `qualified_references` or `full_references` for clarity?

also, should this work for nested references too, like `current.package.name.SomeMessage.SomeSubMessage`?

mvdan (changes requested):
Thanks! Some thoughts below.

hiporox:
@mvdan Thanks for the feedback. I have addressed the comments you had and updated the PR. It now checks against the full name and the short name in `resolveTopScope` and I added to the test case to include doubly nested messages. Let me know what you think

encoding/protobuf/protobuf_test.go
```diff
@@ -23,12 +23,11 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/google/go-cmp/cmp"
-
 	"cuelang.org/go/cue/ast"
 	"cuelang.org/go/cue/errors"
 	"cuelang.org/go/cue/format"
 	"cuelang.org/go/internal/cuetest"
+	"github.com/google/go-cmp/cmp"
```
mvdan: please don't move this import

encoding/protobuf/parse.go
```diff
@@ -271,32 +271,45 @@ func (p *protoConverter) resolve(pos scanner.Position, name string, options []*p
...
-					i = len(name)
-				}
-				expr = ast.NewSel(expr, "#"+name[:i])
-			}
-			ast.SetPos(expr, p.toCUEPos(pos))
-			return expr
 		}
+		if m, ok := p.scope[0][curName]; ok {
+			return p.resolveTopScopeHelper(pos, m.cue, m.pkg, name, i)
+		}
```
mvdan, line 298: you don't need such a large refactor. you can just replace the original line

```
if m, ok := p.scope[0][name[:i]]; ok {
```

with something like

```
curName := name[:i]
if local, ok := strings.CutPrefix(curName, p.protoPkg + "."); ok {
    curName = local
}
if m, ok := p.scope[0][curName]; ok {
```

and you also avoid needing a new method.

mvdan (approved):

encoding/protobuf/protobuf_test.go
hiporox: Sorry, that was probably the auto formatter, will fix

encoding/protobuf/parse.go
hiporox: Fixed

encoding/protobuf/protobuf_test.go
hiporox: Fixed

mvdan (approved):
Thanks!

mvdan:
Thanks! Imported as: https://review.gerrithub.io/c/cue-lang/cue/+/1221581

By the way, I'm curious how you're using CUE, given that you're clearly a relatively advanced user. Are you on our Slack or Discord spaces, so that we could briefly chat next week?

hiporox:
I started using it for my job to define validated and well formatted json to pass into code generators. This is my first time using CUE actually, but I love type systems and it was pretty intuitive coming from a background in strongly typed languages. I tried joining the slack, but it is private. I am on the discord. Same username there as on Github. Would be happy to have a brief chat next week. You can DM me to set up a time

PR #4085, opened 2025-09-12
Update bug_report.md

1007chenbaojun:
yaml with cue are not right


mvdan:
Sorry if there was any confusion, but the template is not for you to send a PR against, it's a template for bug report issues.

PR #4118, opened 2025-10-06
cmd/cue: add --outfile flag to `get go`

### LINKED ISSUE #4108: cmd/cue: Add `--skip-imports`, `--keep-empty`, `--out` to `get go`
As an advanced User relying on Bazel, I want to generate schemas for each packages separately. As Bazel requires to know which files will be created I want to generate a `.cue` file for every input file even when it would be empty. Additionally I do have to explicity tell the command where to write the output to, since everything would run in a sandbox with very tight permissions.

A `--skip-imports`, `--keep-empty` and `--out` flag would be the easiest solution.

Numerous things. Generating empty files to overwrite, trying to guess which files will be generated, deleting unused files. Sadly the `--out` support is requires as I can't find a way to control it in another way.

**Additional context**
This is the patch I am currently using
```patch
diff --git a/cmd/cue/cmd/get_go.go b/cmd/cue/cmd/get_go.go
index 2a81dbdd..9e0c8250 100644
--- a/cmd/cue/cmd/get_go.go
+++ b/cmd/cue/cmd/get_go.go
@@ -45,6 +45,16 @@ import (
 	"cuelang.org/go/internal"
 )
 
+func init() {
+	// Add `go` to PATH for x/go/packages.
+	if goRoot, ok := os.LookupEnv("GOROOT"); ok {
+		goRoot, _ = filepath.Abs(goRoot)
+		goBin := filepath.Join(goRoot, "bin")
+		os.Setenv("GOROOT", goRoot)
+		os.Setenv("PATH", goBin)
+	}
+}
+
 // TODO:
 // Document:
 // - Use ast package.
@@ -216,14 +226,25 @@ restrictive enum interpretation of #Switch remains.
 	cmd.Flags().Bool(string(flagLocal), false,
 		"generates files in the main module locally")
 
+	cmd.Flags().Bool(string(flagSkipImports), false,
+		"generates files without generating recursive through imports")
+
+	cmd.Flags().Bool(string(flagKeepEmpty), false,
+		"generates all files even if they would be empty")
+
+	cmd.Flags().String(string(flagOut), "",
+		"The directory to write the generated files to")
+
 	cmd.Flags().StringP(string(flagPackage), "p", "", "package name for generated CUE files")
 
 	return cmd
 }
 
 const (
-	flagExclude flagName = "exclude"
-	flagLocal   flagName = "local"
+	flagExclude     flagName = "exclude"
+	flagLocal       flagName = "local"
+	flagSkipImports flagName = "skip-imports"
+	flagKeepEmpty   flagName = "keep-empty"
 )
 
 func (e *extractor) initExclusions(str string) {
@@ -247,8 +268,9 @@ func (e *extractor) filter(name string) bool {
 type extractor struct {
 	cmd *Command
 
-	allPkgs map[string]*packages.Package
-	done    map[string]bool
+	allPkgs  map[string]*packages.Package
+	done     map[string]bool
+	rootPkgs map[string]bool
 
 	// per package
 	pkg      *packages.Package
@@ -406,9 +428,11 @@ func extract(cmd *Command, args []string) error {
 	e.initExclusions(flagExclude.String(cmd))
 
 	e.done = map[string]bool{}
+	e.rootPkgs = map[string]bool{}
 
 	for _, p := range pkgs {
 		e.done[p.PkgPath] = true
+		e.rootPkgs[p.PkgPath] = true
 		e.addPackage(p)
 	}
 
@@ -463,6 +487,9 @@ func (e *extractor) extractPkg(root string, p *packages.Package) error {
 
 	pkg := p.PkgPath
 	dir := filepath.Join(load.GenPath(root), filepath.FromSlash(pkg))
+	if out := flagOut.String(e.cmd); out != "" {
+		dir = out
+	}
 
 	isMain := flagLocal.Bool(e.cmd) && p.Module != nil && p.Module.Main
 	if isMain {
@@ -513,7 +540,11 @@ func (e *extractor) extractPkg(root string, p *packages.Package) error {
 			}
 		}
 
-		if len(decls) == 0 && f.Doc == nil {
+		if !e.rootPkgs[p.PkgPath] && flagSkipImports.Bool(e.cmd) {
+			continue
+		}
+
+		if len(decls) == 0 && f.Doc == nil && !flagKeepEmpty.Bool(e.cmd) {
 			continue
 		}
 
```

fionera:
This introduces a way to write generated cue schemas to a given file or stdout. It joins all declarations into one giant file and preserves all package comments.

Closes #4108


cmd/cue/cmd/get_go.go
```diff
@@ -218,6 +219,8 @@ restrictive enum interpretation of #Switch remains.
 
 	cmd.Flags().StringP(string(flagPackage), "p", "", "package name for generated CUE files")
 
+	cmd.Flags().String(string(flagOutFile), "", "write generated CUE schema to the given file or stdout (only allows for one package)")
```
mvdan:
```suggestion
	cmd.Flags().String(string(flagOutFile), "", "generate one CUE file for a single Go package")
```

cmd/cue/cmd/get_go.go
```diff
@@ -398,9 +407,14 @@ func extract(cmd *Command, args []string) error {
 
 	for _, p := range pkgs {
 		e.done[p.PkgPath] = true
+		e.rootPkgs[p.PkgPath] = true
 		e.addPackage(p)
 	}
 
+	if flagOutFile.IsSet(cmd) && len(e.rootPkgs) > 1 {
```
mvdan: check this earlier, via `if flagOutFile.IsSet(cmd) && len(pkgs) != 1` (also note that we want `!= 1` rather than `> 1`)

cmd/cue/cmd/get_go.go
```diff
@@ -883,7 +929,7 @@ func makeDoc(g *ast.CommentGroup, isDoc bool) *cueast.CommentGroup {
 		// The parser has given us exactly the comment text.
 		switch c[1] {
 		case '/':
-			//-style comment (no newline at the end)
+			// -style comment (no newline at the end)
```
mvdan: undo this; we really meant "comments in the double slash style", hence no space. I guess we could have `// //-style comment` but that doesn't seem strictly better

cmd/cue/cmd/get_go.go
```diff
@@ -439,6 +453,10 @@ func (e *extractor) extractPkg(root string, p *packages.Package) error {
 
 	e.recordTypeInfo(p)
 
+	if !e.rootPkgs[p.PkgPath] && flagOutFile.IsSet(e.cmd) {
```
mvdan: rather than this (and needing rootPkgs) you could avoid recursing into `usedPkgs` when the `--outfile` flag is set.

cmd/cue/cmd/get_go.go
```diff
@@ -530,16 +563,29 @@ func (e *extractor) extractPkg(root string, p *packages.Package) error {
...
+		if filePath == "-" {
+			_, err = os.Stdout.Write(b)
+			if err != nil {
+				return err
+			}
+		} else {
+			err = os.WriteFile(filePath, b, 0666)
+			if err != nil {
+				return err
+			}
```
mvdan: nit: might as well move the error check outside of the if/else

cmd/cue/cmd/get_go.go
```diff
@@ -530,16 +563,29 @@ func (e *extractor) extractPkg(root string, p *packages.Package) error {
 		}
 
 		file := filepath.Base(p.CompiledGoFiles[i])
-
 		file = strings.Replace(file, ".go", "_go", 1)
 		file += "_gen.cue"
+		filePath := filepath.Join(dir, file)
```
mvdan: I'd put this in an "else" for flagOutFile.IsSet below, to not give the impression that it's always used

cmd/cue/cmd/get_go.go
```diff
@@ -494,25 +528,24 @@ func (e *extractor) extractPkg(root string, p *packages.Package) error {
...
 
-		pName := flagPackage.String(e.cmd)
-		if pName == "" {
-			pName = p.Name
-		}
+		addDoc(f.Doc, cuePkg)
 
-		pkg := &cueast.Package{Name: e.ident(pName, false)}
-		addDoc(f.Doc, pkg)
+		if flagOutFile.IsSet(e.cmd) && i != len(p.Syntax)-1 {
```
mvdan, line 542: pretty funky. the logic is not particularly easy to follow. can we add more comments? perhaps one jsut above the main for loop, explaining how the regular and "outfile" modes differ.

mvdan (commented):
largely SGTM but needs a test. for example, copy `get_go_vendor.txtar` into `get_go_outfile.txtar` and adapt as you see fit. ideally we test the multi-file package scenario as well.

cmd/cue/cmd/get_go.go
fionera: Yeah. The logic is kinda ugly but I stared the last two hours on this and didn't found a nice way to structure this that wouldn't copy all the logic

mvdan: Nothing obvious comes to mind, the code as is is fine - it just needs a bit of a summary, like "in outfile mode, we accumulate into a single cue file and we write when we're done with the last go file".

cmd/cue/cmd/get_go.go
fionera: My IDEA just doesn't want this :sweat: I added the additonal // so its correct and it doesn't complain about the technically correct warning regarding comments.

fionera:
Ah forgot the test, ignore the request for review. Sorry ^^'

fionera:
Now with added test :smile:

cmd/cue/cmd/get_go.go
```diff
@@ -473,7 +486,25 @@ func (e *extractor) extractPkg(root string, p *packages.Package) error {
...
+		pName = p.Name
+	}
+
+	var (
+		decls  []cueast.Decl
+		cuePkg *cueast.Package
+	)
 	for i, f := range p.Syntax {
+		// If we are in the first iteration we need to set cuePkg.
+		// Af we want to join everything into the same file,
```
mvdan: typo: "As"?

cmd/cue/cmd/get_go.go
```diff
@@ -473,7 +486,25 @@ func (e *extractor) extractPkg(root string, p *packages.Package) error {
...
+		// we have to keep the same cuePkg and declarations around
+		// for all iterations, else reset.
```
mvdan: I'd move this comment above the for loop, and also explain what's happening with writing only the last (joined) file. Something like...

```
// By default, we output one CUE file for each Go file as long as there's anything to generate.
// When --outfile is used, we want to generate exactly one CUE file for an entire Go package,
// so we instead keep joining declarations until we reach the last Go file, where we then write.
```

cmd/cue/cmd/testdata/script/get_go_outfile.txtar
```diff
@@ -0,0 +1,44 @@
...
+cmp outfile.cue x_go_gen.cue.golden
+
+-- go.mod --
+module mod.test
+
+go 1.21
+-- cue.mod/module.cue --
+module: "mod.test"
+language: version: "v0.9.0"
+-- x/x.go --
```
mvdan, line 24: best to include multiple files in `package x`, to make sure that the joining of declarations is tested.

cmd/cue/cmd/get_go.go
```diff
@@ -386,6 +393,10 @@ func extract(cmd *Command, args []string) error {
 		return ErrPrintedError
 	}
 
+	if flagOutFile.IsSet(cmd) && len(pkgs) != 1 {
+		return errors.New("--outfile only allows for one package to be specified")
+	}
```
mvdan, line 398: I'd test these edge cases in the testscript too, like...

```
! exec cue get go --outfile out.cue --local ./x
stderr 'mutually exclusive'
! exec cue get go --outfile out.cue ./x ./y
stderr 'only allows for one package'
```

cmd/cue/cmd/get_go.go
```diff
@@ -549,11 +591,13 @@ func (e *extractor) extractPkg(root string, p *packages.Package) error {
...
-			e.done[path] = true
-			if err := e.extractPkg(root, e.allPkgs[path]); err != nil {
-				return err
+	if !flagOutFile.IsSet(e.cmd) {
+		for pkgPath := range e.usedPkgs {
+			if !e.done[pkgPath] {
+				e.done[pkgPath] = true
+				if err := e.extractPkg(root, e.allPkgs[pkgPath]); err != nil {
+					return err
+				}
```
mvdan:
```suggestion
	for pkgPath := range e.usedPkgs {
		if !flagOutFile.IsSet(e.cmd) && !e.done[pkgPath] {
			e.done[pkgPath] = true
			if err := e.extractPkg(root, e.allPkgs[pkgPath]); err != nil {
				return err
```

mvdan: there's a negligible performance difference either way, so there's no need to add a level of nesting here

cmd/cue/cmd/testdata/script/get_go_outfile.txtar
```diff
@@ -0,0 +1,44 @@
+# Issue 4108
+
+exec cue get go --outfile outfile.cue ./x
+
+# We can only compare the generated file to the expected golden file,
+# as all imports are skipped and only the given package is generated.
+cmp outfile.cue x_go_gen.cue.golden
+
```
mvdan, line 8: perhaps test that we didn't generate any of the usual files too, like...

```
! exists cue.mod/gen/ x/x_go_gen.cue y/y_go_gen.cue
```

mvdan (approved):
Thanks!

cmd/cue/cmd/get_go.go
mvdan: (ignore given my suggested rewrite below)

cmd/cue/cmd/get_go.go
fionera: Actually just found it harder to follow the logic, but fine for me :D

mvdan (approved):
Thanks! Imported to Gerrit here: https://cue.gerrithub.io/c/cue-lang/cue/+/1224215

You can send patches in the future to Gerrit directly, which will avoid this dance. See: https://github.com/cue-lang/cue/blob/master/CONTRIBUTING.md#preparing-for-gerrithub-cl-contributions

PR #4191, opened 2025-11-22
cmd/cue: detect kubernetes semantics for `get go`

fionera:
Kubernetes types with either +optional or omitempty, are expected to not have a null disjunction. By checking for their annotations we can switch to a kubernetes semantic mode and not generate them.


cmd/cue/cmd/get_go.go
```diff
@@ -444,9 +445,39 @@ func (e *extractor) recordTypeInfo(p *packages.Package) {
 	}
 }
 
+type importSematic int
```
mvdan: given that we only need "default" vs "k8s" for now, I think just a boolean is fine. we can always expand to an enum type in the future if needed. I suspect/hope that we won't need more modes anytime soon.

cmd/cue/cmd/get_go.go
```diff
@@ -444,9 +445,39 @@ func (e *extractor) recordTypeInfo(p *packages.Package) {
...
 func (e *extractor) extractPkg(root string, p *packages.Package) error {
 	e.pkg = p
-	e.logf("--- Package %s", p.PkgPath)
+	e.semantic = semanticDefault
+
+	for _, f := range p.Syntax {
+		for _, c := range f.Comments {
+			for _, c := range c.List {
+				switch c.Text {
+				case "// +k8s:openapi-gen=true":
```
mvdan: personally I might do a `strings.Contains(c.Text(), "+k8s:openapi-gen=true")` - it's simpler, it's less likely to cause weird false negatives with whitespace, and I doubt it would cause any false positives ever.

cmd/cue/cmd/get_go.go
```diff
@@ -1209,10 +1240,13 @@ func (e *extractor) makeType(typ types.Type) (result cueast.Expr) {
 
 		return result
 	case *types.Pointer:
+		if e.semantic == semanticKubernetes && kind == optional {
```
mvdan: make a comment here, like "in k8s, a field is only nullable if explicitly marked as such"

cmd/cue/cmd/testdata/script/get_go_types_kubernetes.txtar
```diff
@@ -0,0 +1,601 @@
+# Test that a get go works for kubernetes types,
```
mvdan, line 1: this test file copies all of get_go_types.txtar, which works, but it's very verbose. we only really care about a few edge cases which behave differently in k8s mode, so I'd strongly suggest writing a much smaller test that's focused on that, in 50 lines or so.

mvdan (commented):
SGTM overall, but I think this implementation isn't quite complete - it should support the "optional" and "nullable" field comment attributes that the k8s openapi tool supports. Note that we already support `+optional` (and I'm not sure if that's already compatible with what k8s does), but we don't support `+nullable` at all.

mvdan:
Also, could you split this in two commits as follows:
1) Add the testscript file, showing what the current implementation generates for k8s types
2) Add the code changes, showing how it affects the generated types from the first commit, effectively showing a diff removing `null | ` etc

cmd/cue/cmd/testdata/script/get_go_types_kubernetes.txtar
```diff
@@ -0,0 +1,48 @@
...
+# including those special cases that result in _ and string
+# that are specifically mentioned in help get go.
+
+# cue get go
```
mvdan: nit: remove this line

cmd/cue/cmd/testdata/script/get_go_types_kubernetes.txtar
```diff
@@ -0,0 +1,48 @@
...
+exec cue get go --local ./...
+cmp ./file2_go_gen.cue ./file2_go_gen.cue.golden
+
+-- go.mod --
+module mod.test
+
+go 1.21
+-- cue.mod/module.cue --
+module: "mod.test"
+-- file1.go --
```
mvdan, line 14: call this `doc.go` to be more consistent with upstream?

cmd/cue/cmd/testdata/script/get_go_types_kubernetes.txtar
```diff
@@ -0,0 +1,48 @@
...
+package pkg1
+
+// +k8s:openapi-gen=true
```
mvdan, line 17: I seem to recall that the files we looked at had this +k8s comment _above the package declaration_. This is important, because it means we don't need to scan the comments from the entire file to see if the comment exists or not. I'd move it up here to be consistent with that expectation.

cmd/cue/cmd/testdata/script/get_go_types_kubernetes.txtar
```diff
@@ -0,0 +1,48 @@
...
+
+type Foo struct {
+	Required string `json:"required"`
+	Pointer *string `json:"pointer"`
+	OmitEmptyTagPointer *string `json:"omitEmptyTagPointer,omitempty"`
+	// +optional
+    OptionalCommentTagPointer *string `json:"optionalCommentTagPointer"`
+	// +nullable
+    NullableCommentTagPointer *string `json:"nullableCommentTagPointer"`
+}
```
mvdan, line 31: I'd add something like an `+optional` non-pointer string, for completeness.

Also, is the word "Tag" in the fields above adding anything? I'd personally remove it.

cmd/cue/cmd/get_go.go
```diff
@@ -446,7 +447,16 @@ func (e *extractor) recordTypeInfo(p *packages.Package) {
 
 func (e *extractor) extractPkg(root string, p *packages.Package) error {
 	e.pkg = p
-	e.logf("--- Package %s", p.PkgPath)
+	e.k8sSemantic = false
+
+	for _, f := range p.Syntax {
```
mvdan, line 453: the example from upstream we looked at had the comment in `doc.go`, is that something that we can rely on so that we don't scan all files?

cmd/cue/cmd/get_go.go
```diff
@@ -446,7 +447,16 @@ func (e *extractor) recordTypeInfo(p *packages.Package) {
...
+		for _, c := range f.Comments {
```
mvdan, line 454: you should only scan up to the package declaration line. not just for the sake of performance, but also so that any comments elsewhere in the file don't incorrectly affect our logic here.

basically: loop over f.Comments, but break as soon as `c.Pos() > f.Package`.

cmd/cue/cmd/get_go.go
```diff
@@ -1480,24 +1491,36 @@ func (e *extractor) isInline(tag string) bool {
...
-	if _, ok := f.Type().(*types.Pointer); ok {
-		return true
-	}
-
+func (e *extractor) detectFieldKind(f *types.Var, doc *ast.CommentGroup, tag string) fieldKind {
+	// From k8s docs:
+	// - a map field marked with `+nullable` would accept either `foo: null` or `foo: {}`.
+	// - Using the `+optional` or the `omitempty` tag means that the field is optional.
+	//   Note that for backward compatibility, any field that has the `omitempty` struct
+	//   tag, and is not explicitly marked as `+required`, will be considered to be optional.
```
mvdan, line 1501: this doesn't have to happen in this PR, but I'd leave a TODO about supporting `+required`, as it seems to me like we should support it for complete compatibility of these rules here.

cmd/cue/cmd/get_go.go
```diff
@@ -1480,24 +1491,36 @@ func (e *extractor) isInline(tag string) bool {
...
 	for line := range strings.SplitSeq(doc.Text(), "\n") {
 		before, _, _ := strings.Cut(strings.TrimSpace(line), "=")
-		if before == "+optional" {
-			return true
+		switch before {
+		case "+nullable":
+			return regular
+		case "+optional":
+			return optional
```
mvdan, line 1508: I see from upstream examples that they support key-values like `+featureGate=CRDValidationRatcheting`; do they support e.g. `+optional=false`?

mvdan (approved):
LGTM - just a few last nits and I'm happy to import into Gerrit and merge

cmd/cue/cmd/testdata/script/get_go_types_kubernetes.txtar
fionera: Sounds reasonable, tho I found a few things where they are on the struct itself
```
// +k8s:deepcopy-gen=package

package apps
```

```
// +protobuf=true
// +protobuf.options.(gogoproto.goproto_stringer)=false
// +k8s:openapi-gen=true
// +k8s:openapi-model-package=io.k8s.apimachinery.pkg.util.intstr
type IntOrString struct {
	Type   Type   `protobuf:"varint,1,opt,name=type,casttype=Type"`
	IntVal int32  `protobuf:"varint,2,opt,name=intVal"`
	StrVal string `protobuf:"bytes,3,opt,name=strVal"`
}
```

fionera: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/intstr/intstr.go

cmd/cue/cmd/get_go.go
fionera: Sadly no, see comment above

cmd/cue/cmd/get_go.go
fionera: I think we have to go over all comments, as stated above. I added a break tho.

cmd/cue/cmd/get_go.go
fionera: No, optional doesnt have any arguments

cmd/cue/cmd/get_go.go
fionera: There is a TODO at https://github.com/cue-lang/cue/blob/master/cmd/cue/cmd/get_go.go#L1067

cmd/cue/cmd/testdata/script/get_go_types_kubernetes.txtar
fionera: Sadly they are not always called doc.go :(

cmd/cue/cmd/testdata/script/get_go_types_kubernetes.txtar
mvdan: ack, thank you

cmd/cue/cmd/testdata/script/get_go_types_kubernetes.txtar
```diff
@@ -0,0 +1,52 @@
+# Test that a get go works for kubernetes types,
+# including those special cases that result in _ and string
+# that are specifically mentioned in help get go.
+
+exec cue get go --local ./...
```
mvdan, line 5: nit: replace `./...` with `.`

cmd/cue/cmd/testdata/script/get_go_types_kubernetes.txtar
```diff
@@ -0,0 +1,52 @@
...
+type Foo struct {
+	Required string `json:"required"`
+	Pointer *string `json:"pointer"`
+	OmitEmptyPointer *string `json:"omitEmptyPointer,omitempty"`
+	// +optional
+    OptionalComment string `json:"optionalComment"`
+	// +optional
+    OptionalCommentPointer *string `json:"optionalCommentPointer"`
+	// +nullable
+    NullableCommentPointer *string `json:"nullableCommentPointer"`
```
mvdan, line 30: I just realised - we need to include a field which is both optional and nullable. it should generate as...

```
foo?: null | string
```

cmd/cue/cmd/testdata/script/get_go_types_kubernetes.txtar
```diff
@@ -0,0 +1,52 @@
...
+module: "mod.test"
+-- file1.go --
+package pkg1
+
+// +k8s:openapi-gen=true
+-- file2.go --
+package pkg1
+
+type Foo struct {
+	Required string `json:"required"`
```
mvdan, line 22: For the sake of completeness in this test, I'd rename this to "Regular", and add a "Required" field with the `+required` comment.

cmd/cue/cmd/get_go.go
```diff
@@ -1480,24 +1493,36 @@ func (e *extractor) isInline(tag string) bool {
...
 		}
```
mvdan, line 1509: I _think_ that the way this loops over the doc lines, and returns at the first hit, will not work well when a field is marked as BOTH optional and nullable. You might need to make fieldKind a bit set, so that we can track that a field was marked as either or both optional/nullable?

If this is not common and you want to leave this tweak for a future PR, that's fine, but then we'd need a TODO.

mvdan (approved):
Thanks! LGTM with some last nits.

When you're ready, could you please send these for a final review and test/merge on Gerrit? The way we import PRs to Gerrit doesn't really support multi-commit PRs unless we squash them, and I don't want to squash here.

cmd/cue/cmd/testdata/script/get_go_types_kubernetes.txtar
fionera: Turns out that should be the default behavior:

```
Optional fields have the following properties:

They have the +optional comment tag in Go.
They are a pointer type in the Go definition (e.g. AwesomeFlag *SomeFlag) or have a built-in nil value (e.g. maps and slices).
They are marked with the omitempty json struct tag in the Go definition.
The API server should allow POSTing and PUTing a resource with this field unset.

```

cmd/cue/cmd/get_go.go
fionera: I actually changed the whole PR :sweat_smile: I am gonna close this one and push it to gerrit as its probably a better review platform

PR #4199, opened 2025-11-27
encoding/yaml: add support for YAML tags via @yaml(tag) attribute

### LINKED ISSUE #2316: encoding/yaml: Support for encoding tags

There are some situations where a yaml tag is required. `(encoding/yaml).Marshal` does not support encoding tags and makes this impossible.

It looks like the authors of the original encoder were aware this may be desirable.

https://github.com/cue-lang/cue/blob/256f1f9ee1611636fef85446ff9ea3b7d4f6b7d0/internal/encoding/yaml/encode.go#L107

Support encoding tags with an attribute.

```cue
import "encoding/yaml"

yaml.Marshal({
    some_key: "some value" @yaml.Tag("some tag")
})
```

Not sure.

**Additional context**

Maybe:
- https://github.com/cue-lang/cue/issues/422
- https://github.com/cue-lang/cue/issues/1070

jonas-meyer:
This adds support for encoding YAML tags by specifying them in @yaml(tag="...") attributes on CUE fields. The tag value is applied to the encoded YAML node.

This enables generating YAML files that use custom tags, such as those required by authentik blueprints (!KeyOf, !Env, !Find, etc.) and other YAML-based tools that rely on tags for special processing.

Tags can be applied to scalars, sequences, and mappings. The implementation extracts tags from field attributes in the AST and sets them on the corresponding yaml.Node during encoding.

Fixes #2316

### DIFF
--- a/internal/encoding/yaml/encode.go
+++ b/internal/encoding/yaml/encode.go
@@ -234,6 +234,28 @@ func encodeExprs(exprs []ast.Expr) (n *yaml.Node, err error) {
 	return n, nil
 }
 
+// extractYAMLTag looks for @yaml(,tag="...") attribute and returns the tag value.
+// Returns an empty string if no @yaml attribute or no tag argument is found.
+// Returns an error if the attribute is malformed.
+func extractYAMLTag(attrs []*ast.Attribute) (string, error) {
+	for _, attr := range attrs {
+		key, body := attr.Split()
+		if key != "yaml" {
+			continue
+		}
+		parsed := internal.ParseAttrBody(attr.Pos(), body)
+		if parsed.Err != nil {
+			return "", parsed.Err
+		}
+		if val, found, err := parsed.Lookup(1, "tag"); err != nil {
+			return "", err
+		} else if found {
+			return val, nil
+		}
+	}
+	return "", nil
+}
+
 // encodeDecls converts a sequence of declarations to a value. If it encounters
 // an embedded value, it will return this expression. This is more relaxed for
 // structs than is currently allowed for CUE, but the expectation is that this
@@ -289,6 +311,15 @@ func encodeDecls(decls []ast.Decl) (n *yaml.Node, err error) {
 			if err != nil {
 				return nil, err
 			}
+
+			yamlTag, err := extractYAMLTag(x.Attrs)
+			if err != nil {
+				return nil, err
+			}
+			if yamlTag != "" {
+				value.Tag = yamlTag
+			}
+
 			lastHead = label
 			lastFoot = value
 			addDocs(x, label, value)
--- a/internal/encoding/yaml/encode_test.go
+++ b/internal/encoding/yaml/encode_test.go
@@ -243,6 +243,78 @@ route:
   receiver: pager
   group_by: [alertname, cluster]
 		`,
+	}, {
+		name: "yaml_tag_scalar",
+		in: `
+		key: "value" @yaml(,tag="!Custom")
+		env: "VAR_NAME" @yaml(,tag="!Env")
+		`,
+		out: `
+key: !Custom value
+env: !Env VAR_NAME
+		`,
+	}, {
+		name: "yaml_tag_sequence",
+		in: `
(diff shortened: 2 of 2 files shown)


freakhill:
little bump. this would be great in order to use the atmos terraform generation/orchestration tool with cue.

jonas-meyer:
Don't like bumping but would love for this to get merged! @mvdan

mvdan:
All good, thanks for the bump. I'm going through regressions, bug reports, and patches as quickly as I can :)

internal/encoding/yaml/encode.go
```diff
@@ -234,6 +234,27 @@ func encodeExprs(exprs []ast.Expr) (n *yaml.Node, err error) {
...
+// Returns an error if the attribute is malformed.
+func extractYAMLTag(attrs []*ast.Attribute) (string, error) {
+	for _, attr := range attrs {
+		key, body := attr.Split()
+		if key == "yaml" {
+			parsed := internal.ParseAttrBody(attr.Pos(), body)
+			if parsed.Err != nil {
+				return "", parsed.Err
+			}
+			if val, found, err := parsed.Lookup(0, "tag"); err != nil {
```
mvdan:
```suggestion
			if val, found, err := parsed.Lookup(1, "tag"); err != nil {
```

This is akin to what we do in other encodings like Go, and allows us to later support `@yaml(name)` to specify a name to encode/decode with rather than the one inferred from a CUE field.

And the test cases below should replace e.g. `@yaml(tag="!Find")` with `@yaml(,tag="!Find")` accordingly.

internal/encoding/yaml/encode.go
```diff
@@ -234,6 +234,27 @@ func encodeExprs(exprs []ast.Expr) (n *yaml.Node, err error) {
 	return n, nil
 }
 
+// extractYAMLTag looks for @yaml(tag="...") attribute and returns the tag value.
+// Returns an empty string if no @yaml attribute or no tag argument is found.
+// Returns an error if the attribute is malformed.
+func extractYAMLTag(attrs []*ast.Attribute) (string, error) {
+	for _, attr := range attrs {
+		key, body := attr.Split()
+		if key == "yaml" {
```
mvdan: nit: invert this check and early continue to unindent the rest

internal/encoding/yaml/encode.go
```diff
@@ -289,6 +310,15 @@ func encodeDecls(decls []ast.Decl) (n *yaml.Node, err error) {
...
 				return nil, err
 			}
+
+			yamlTag, err := extractYAMLTag(x.Attrs)
+			if err != nil {
+				return nil, err
+			}
+			if yamlTag != "" {
+				value.Tag = yamlTag
+			}
```
mvdan, line 321: `value.Tag` is a string, so can we not just unconditionally do `value.Tag = yamlTag` here?

internal/encoding/yaml/encode_test.go
```diff
@@ -243,6 +243,70 @@ route:
...
+		out: `
+plain: no-tag
+tagged: !Custom has-tag
+nested:
+  field: !Nested value
+		`,
+	}, {
+		name: "yaml_tag_verbatim",
+		in: `
+		custom: "value" @yaml(tag="!<tag:example.com,2000:app/foo>")
```
mvdan: I seem to recall that YAML tags can also contain e.g. HTTPS URLs, can you please add a test case for that too?

mvdan (approved):
Thanks! LGTM with some minor suggestions.

And thanks for being patient - with the holidays and us attending two conferences last month, we were a bit busy :)

internal/encoding/yaml/encode.go
jonas-meyer: An empty yamlTag would overwrite tags already set by the encoder, e.g. !!binary for byte literals. Kept the guard to avoid that.

mvdan (approved):
Thanks!

mvdan:
Importing to Gerrit here: https://cue.gerrithub.io/c/cue-lang/cue/+/1231452

Issue #2956, opened 2024-03-18
mod: implement replace directive


rogpeppe:
Go's module system provides a `replace` directive to allow replacing
one module by another module or by a module that is present in a local
directory.

CUE's modules should implement that too.

morlay:
When `replace directive` supported. 

We could maintain cue modules in one monorepo.

**code layout**:

```txtar
-- a/cue.mod/module.cue --
module: "ghcr.io/morlay/a@v0"
language: {
        version: "v0.8.0"
}
-- a/x.cue --
package a

#Version: "v0.0.0"
-- b/cue.mod/module.cue --
module: "ghcr.io/morlay/b@v0"
language: {
        version: "v0.8.0"
}
deps: {
        "ghcr.io/morlay/a@v0": {
                "v":       "v0.0.0"
                "replace": "../a"
        }
}
-- b/x.cue --
package b

import (
        "ghcr.io/morlay/a"
)

deps: {
        a: "a.#Version"
}
-- c/cue.mod/module.cue --
module: "ghcr.io/morlay/c@v0"
language: {
        version: "v0.8.0"
}
deps: {
        "ghcr.io/morlay/a@v0": {
                "v":       "v0.0.0"
                "replace": "../a"
        }
}
-- c/x.cue --
package c

import (
        "ghcr.io/morlay/a"
)

deps: {
        a: "a.#Version"
}
```

**module deps**:

```mermaid
graph TD
a["ghcr.io/morlay/a"]
b["ghcr.io/morlay/b"]
c["ghcr.io/morlay/c"]

a --> b;
a --> c;
```

Now we start to publish all cue modules.

First

```shell
cue publish ghcr.io/morlay/a v0.1.0
```

Then
```shell
cue publish ghcr.io/morlay/b v0.1.0
cue publish ghcr.io/morlay/c v0.1.0
```

**but version of `ghcr.io/morlay/a` already `v0.1.0`**, 
we have to change the version of  `ghcr.io/morlay/a` in `ghcr.io/morlay/b` and  `ghcr.io/morlay/c`, before publish.

So. may consider added optional `version?: string` in  `module.cue`,
Then`cue publish ./a` could use the `module` and `version` form `module.cue`
And in the case above, `cue publish ./a` and `cue publish ./c` could resolve the correct version of `ghcr.io/morlay/a`.

rogpeppe:
[...]

PR #4214, opened 2025-12-17
mod: add local module replacement support with robust path resolution

rawkode:
Add support for local path replacements in module.cue files, similar to Go's replace directive. This allows replacing module dependencies with local directories using paths like "./local-dep" or "../sibling".

First pass at #2956

Key changes:
- Add modreplace package for handling local path replacements
- Add localreg.go to wrap Registry with replacement support
- Update schema.cue to allow empty version for replace-only deps
- Compute absolute paths at creation time instead of using os.Getwd()
- Add comprehensive tests including script tests for various scenarios

The path resolution now requires either:
- A filesystem implementing module.OSRootFS, or
- An absolute directory path

This eliminates fragile reliance on os.Getwd() which could fail in containers, tests, or when the working directory changes.


internal/mod/modfiledata/modfile.go
```diff
@@ -193,13 +229,68 @@ func (mf *File) init(strict bool) error {
...
+		return Replacement{}, fmt.Errorf("absolute path replacement %q not allowed; use relative path starting with ./ or ../", replace)
+	}
+	// Check for Windows-style absolute paths (C:\foo or C:/foo) - first char must be a letter
+	if len(replace) >= 3 && unicode.IsLetter(rune(replace[0])) && replace[1] == ':' && (replace[2] == '/' || replace[2] == '\\') {
+		return Replacement{}, fmt.Errorf("absolute path replacement %q not allowed; use relative path starting with ./ or ../", replace)
+	}
+	// Reject UNC paths (\\server\share or //server/share)
+	if len(replace) >= 2 && ((replace[0] == '\\' && replace[1] == '\\') || (replace[0] == '/' && replace[1] == '/')) {
+		return Replacement{}, fmt.Errorf("UNC path replacement %q not allowed; use relative path starting with ./ or ../", replace)
+	}
```
mvdan: we have Unix and Windows path support via pkg/path - use those APIs like so:

```
	if cuepath.IsAbs(s, cuepath.Windows) || cuepath.IsAbs(s, cuepath.Unix) {
```

internal/mod/modload/tidy.go
```diff
@@ -127,7 +138,24 @@ func equalRequirements(rs0, rs1 *modrequirements.Requirements) bool {
 	// Note that we clone the slice to not modify rs1's internal slice in-place.
 	rs1RootMods := slices.DeleteFunc(slices.Clone(rs1.RootModules()), module.Version.IsLocal)
 	return slices.Equal(rs0.RootModules(), rs1RootMods) &&
-		maps.Equal(rs0.DefaultMajorVersions(), rs1.DefaultMajorVersions())
+		maps.Equal(rs0.DefaultMajorVersions(), rs1.DefaultMajorVersions()) &&
+		equalReplacements(rs0.Replacements(), rs1.Replacements())
```
mvdan: why can't you use `maps.Equal` here?

mod/modfile/schema.cue
```diff
@@ -80,16 +80,22 @@ versions: "v0.9.0-alpha.0": {
 
 		#Dep: {
 			// v indicates the minimum required version of the module. This can
-			// be null if the version is unknown and the module entry is only
-			// present to be replaced.
-			v!: #Semver | null
+			// be null or empty when the module entry is only present to specify
+			// a local path replacement (version-independent replacement).
+			v!: #Semver | null | ""
```
mvdan: it's unclear to me why we need to start allowing the empty string; "null" seems like the right value for a replaced module.

mvdan (commented):
Did a quick pass and spotted a few issues.

Overall I notice that you're adding ~900 lines of tests, mostly unit tests, to support ~400 lines of added code. This seems like a case where the AI has gone into overdrive and covered every possible edge case. Which is not a bad thing per se, but we tend to lean towards covering the most common good and bad paths via integration tests (like testscript), and we only use unit tests where necessary, e.g. rare edge cases or for what doesn't fit testscript.

So my intuition is that the 900 lines of tests should be reduced by about half. For example:
* `modreplace_local_commands.txtar` can be joined with `modreplace_local.txtar`
* `modreplace_local_nested.txtar` doesn't seem to test anything new over `modreplace_local.txtar` - they both replace into a module nested underneath the main one
* I'm not sure that `internal/mod/modreplace/local_test.go` adds much over e.g. `modreplace_local_missing.txtar`, at least in terms of realistic scenarios from the CLI

I also wonder if a new `modreplace` package is warranted - I think it's a bit premature for just 150 lines of code.

internal/mod/modfiledata/modfile_test.go
```diff
@@ -0,0 +1,218 @@
...
+			replace: "example.com/bar",
+			wantErr: "invalid replacement",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			repl, err := parseReplacement(tt.oldPath, tt.replace, tt.strict)
+
+			if tt.wantErr != "" {
```
rogpeppe: FWIW I'd tend to write this kind of check logic with the qt package - it's considerably less verbose:

```
if tt.wantErr != "" {
    qt.Assert(t, qt.ErrorMatches(err, tt.wantErr))
    return
}
````

rawkode:
I have hopefully fixed this up.

If you don't think this is on the right path, I am happy to close this.

rogpeppe:
@rawkode I'd like to post a comment to apologise once again for the slowness of turnaround on this review. It's still very much on our radar and I intend to get onto it once I'm back from Shetland madness next week!

mvdan:
Thanks @rawkode for contributing again :) As I'm sure you've seen Roger post in #2956, we've now published a proposal (https://github.com/cue-lang/cue/discussions/4389) and it's implemented in the latest alpha, if you'd like to give it a go and provide feedback in the proposal thread.

PR #4222, opened 2026-01-02
cue/load: add support for loading from io/fs.FS

pskry:
This PR adds support for loading CUE packages and modules from a virtual
filesystem provided via `load.Config.FS`.

This enables embedding CUE modules in Go binaries and similar use cases where
access to the host filesystem is not desired or not possible.

When `Config.FS` is set, all filesystem operations performed by the loader
(stat, read, open, directory walking, and encoding reads) are routed through
`io/fs.FS` instead of the host operating system filesystem. This enables
loading CUE code from sources such as `embed.FS`, `fstest.MapFS`, or other
virtual filesystems, without accessing the host filesystem.

The existing behavior is preserved when `Config.FS` is nil.

#### Design notes

- A small internal abstraction (`loadFS`) centralizes filesystem access and
  normalizes host and virtual filesystem behavior.
- When loading from a virtual filesystem, paths are interpreted as absolute
  loader paths rooted at a synthetic root (`/@fs`), avoiding accidental host
  filesystem access.
- Overlays continue to work and must use absolute loader paths when `FS` is set.
- Encoding now receives an injected open function to ensure it does not bypass
  the loader filesystem.

#### Tests

New tests cover:
- Loading a module and packages from `fstest.MapFS`
- Ensuring the host filesystem is not accessed when `FS` is set
- Subdirectory loading via `Config.Dir`
- Overlay behavior with a virtual filesystem
- Ensuring encoding does not fall back to `os.Open`

No public APIs are removed, and existing behavior is unchanged unless
`Config.FS` is explicitly set.

This change is intended to support embedding CUE modules in Go binaries and similar use cases.

### DIFF
--- a/cue/load/config.go
+++ b/cue/load/config.go
@@ -20,6 +20,7 @@ import (
 	"io"
 	"io/fs"
 	"os"
+	"path"
 	"path/filepath"
 	"runtime"
 
@@ -164,6 +165,28 @@ type Config struct {
 	//         in the _ package.
 	Package string
 
+	// FS, if non-nil, provides the filesystem used by the loader for discovering
+	// packages, resolving modules, and reading files.
+	//
+	// When FS is nil, the loader uses the host operating system filesystem
+	// (os.Stat, os.ReadDir, os.Open), which is the default and preserves the
+	// existing behavior.
+	//
+	// When FS is set, all filesystem operations performed by the loader are
+	// routed through FS. In this mode, paths are interpreted as slash-separated
+	// paths within FS, and Config.Dir is treated as a directory inside FS rather
+	// than a host filesystem path.
+	//
+	// The loader does not require FS to support symbolic links. When using FS,
+	// any operations that would rely on os.Lstat semantics on the host filesystem
+	// are performed using fs.Stat instead, consistent with the capabilities of
+	// io/fs.
+	//
+	// FS enables loading CUE packages and modules from virtual or embedded
+	// filesystems (for example, embed.FS or fstest.MapFS) without accessing
+	// the host filesystem.
+	FS fs.FS
+
 	// Dir is the base directory for import path resolution.
 	// For example, it is used to determine the main module,
 	// and rooted import paths starting with "./" are relative to it.
@@ -279,7 +302,7 @@ type Config struct {
 	// An application may supply a custom implementation of ParseFile to change
 	// the effective file contents or the behavior of the parser, or to modify
 	// the syntax tree.
-	ParseFile func(name string, src interface{}, cfg parser.Config) (*ast.File, error)
+	ParseFile func(name string, src any, cfg parser.Config) (*ast.File, error)
 
 	// Overlay provides a mapping of absolute file paths to file contents,
 	// which are overlaid on top of the host operating system when loading files.
@@ -289,6 +312,9 @@ type Config struct {
 	// If an overlaid file does not exist in the host filesystem,
 	// the loader behaves as if the overlaid file exists with its contents,
 	// and that that all of its parent directories exist too.
+	//
+	// When FS is set, overlay paths must be absolute loadFS paths
+	// (that is, rooted at loadFsRoot).
 	Overlay map[string]Source
 
 	// Stdin defines an alternative for os.Stdin for the file "-". When used,
@@ -353,13 +379,22 @@ func addImportQualifier(pkg importPath, name string) (importPath, error) {
 func (c Config) complete() (cfg *Config, err error) {
 	// Ensure [Config.Dir] is a clean and absolute path,
 	// necessary for matching directory prefixes later.
-	if c.Dir == "" {
-		c.Dir, err = os.Getwd()
-		if err != nil {
+	if c.FS == nil {
+		if c.Dir == "" {
+			c.Dir, err = os.Getwd()
+			if err != nil {
+				return nil, err
+			}
+		} else if c.Dir, err = filepath.Abs(c.Dir); err != nil {
 			return nil, err
 		}
-	} else if c.Dir, err = filepath.Abs(c.Dir); err != nil {
-		return nil, err
+	} else {
+		if c.Dir == "" {
+			c.Dir = "."
+		}
+		if !isLoaderAbs(c.Dir) {
+			c.Dir = path.Clean(path.Join(loadFSRoot, c.Dir))
+		}
 	}
 
 	// TODO: we could populate this already with absolute file paths,
--- a/cue/load/fs.go
+++ b/cue/load/fs.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	iofs "io/fs"
 	"os"
+	"path"
 	"path/filepath"
 	"slices"
 	"strings"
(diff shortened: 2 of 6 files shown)


pskry:
Wow, I am sorry to see this failing on windows. I do not have access to a windows machine, so can't test it locally.
I'm pretty sure, this is caused by "/@fs" shenanigans for the virtual FS root.

I hope someone will review this PR anyways, maybe even give some (much needed) windows insight into pathing there.

pskry:
The Windows CI issues turned out to be path canonicalization problems in the overlay/virtual FS boundary and are now resolved. All tests are green across platforms.

This PR should now be ready for review.

The change is intended to enable a fairly common use case (loading CUE modules from embedded or virtual filesystems) while keeping existing behavior unchanged when Config.FS is nil. I hope this turns out to be useful for others embedding or sandboxing CUE.

Feedback very welcome — especially around the overall approach and any edge cases I may have missed.

pskry:
Hi!

Just wanted to gently bump this in case it got lost in the queue.
Happy to adjust the approach or add tests/docs if needed — just let me know.

Thanks for taking a look 🙂

mvdan:
Hi @pskry, thanks for your patch. We're currently quite busy preparing for upcoming conferences, especially https://cfgmgmtcamp.org/ghent2026/, so it's not likely we will get to this until February.

My only recollection is that when @rogpeppe attempted this a couple of years back, he was really struggling with file positions on Windows; `io/fs` uses forward slashes and does not support volume name prefixes for absolute paths, so we have to really be careful that uses of the `cue` tool on Windows continue to provide the correct Windows file paths back to the user.

Did you encounter a similar issue, and if so, how did you approach it?

pskry:
I'm not sure what you are referring to. I'm using CUE exclusively as embedded library, not as an external CLI tool.
The `FS` option I added to `load.Config` is not (AFAIK) exposed to the CLI entrypoint.
For anything Windows I relied on the test suite.

If `FS` is set, CUE does not touch the host-OS **at all* any longer. It routes all filesystem access through the `FS` configured in the loader using a 'virtual' root (which I just called `@fs`). Thus all paths are normalized to forward slashes (not using `filepath.*`, but `path.*` and rooted in `/@fs`).

I would highly appreciate if someone with access to a Windows machine to do a bit of testing/snooping around, since I sadly cannot.

cue/load/fs.go
```diff
@@ -56,17 +57,179 @@ func (f *overlayFile) Mode() iofs.FileMode {
...
+}
+
+func stripFSRoot(path string) string {
+	tr := func(p string) string {
+		p = strings.ReplaceAll(p, `\`, `/`)
+		p = strings.TrimPrefix(p, loadFSRoot)
+		return strings.TrimPrefix(p, `/`)
+	}
+
+	for strings.Contains(path, loadFSRoot) {
```
rogpeppe, line 124: I don't understand this logic - shouldn't the `@fs` prefix _only_ appear at the very start of the path, and always be succeeded by a separator (or end of string) ?

As is, this can go into an infinite loop (try running `stripFSRoot` on the string `foo@fs`, for example)

I'm not entirely convinced we should be translating backslashes either.
Couldn't this function just be something like:

```
func stripFSRoot(path string) string {
    prefix, rest, _ := strings.Cut(path, "/")
    if prefix == loadFSRoot {
       if rest == "" {
            return "."
       }
       return rest
    }
    return path
}
```
?

cue/load/fs.go
```diff
@@ -56,17 +57,179 @@ func (f *overlayFile) Mode() iofs.FileMode {
...
+	return iofs.Stat(l.fs, strip)
+}
+
+func (l loadFS) Lstat(name string) (iofs.FileInfo, error) {
+	if l.fs == nil {
+		return os.Lstat(name)
+	}
+
+	strip := stripFSRoot(name)
+	if strip == "" {
```
rogpeppe, line 149: Why is this logic different from the "no prefix -> use dot" logic of the rest of the methods?

cue/load/fs.go
```diff
@@ -56,17 +57,179 @@ func (f *overlayFile) Mode() iofs.FileMode {
...
+
+	for strings.Contains(path, loadFSRoot) {
+		path = tr(path)
+	}
+
+	return path
+}
+
+func (l loadFS) Stat(name string) (iofs.FileInfo, error) {
+	if l.fs == nil {
```
rogpeppe, line 132: I'm wondering whether, rather than having an if statement in every method, it might be nicer to define `loadFS` as an interface and have two implementations, one for FS and one for OS. Then we'd have something like:

```
func (l fsLoadFS) Stat(name string) (iofs.FileInfo, error) {
	return iofs.Stat(l.fs, stripFSRoot(name))
}

func (l osLoadFS) Stat(name string) (iofs.FileInfo, error) {
	return os.Stat(name)
}
```

cue/load/fs_test.go
```diff
@@ -57,6 +61,262 @@ func TestIOFS(t *testing.T) {
...
+	}
+
+	cfg := &Config{
+		FS:  cfs,
+		Dir: ".",
+	}
+
+	// Trap host FS access
+	old := source.OsOpen
+	source.OsOpen = func(name string) (fs.File, error) {
```
rogpeppe, line 283: It seems like this is the sole reason why we're passing around an open function and defining source.OsOpen as a variable. In general I'd prefer to avoid that kind of pollution of test concerns into the main logic, and the confidence this gives isn't hugely great anyway (the code could potentially just be invoking os.Open directly).

Rather than fake-out OsOpen, how about just using a name that we have good confidence won't be in the local OS filesystem?

cue/load/fs_test.go
```diff
@@ -57,6 +61,262 @@ func TestIOFS(t *testing.T) {
...
+	qt.Assert(t, qt.IsFalse(called))
+}
+
+type assertNoSyntheticFS struct {
+	t *testing.T
+}
+
+func (fsys assertNoSyntheticFS) Open(name string) (fs.File, error) {
+	if strings.HasPrefix(name, loadFSRoot) {
+		fsys.t.Fatalf("fs.FS.Open called with synthetic path %q", name)
```
rogpeppe, line 302: Rather than invoking Fatal in the middle of arbitrary cue/load logic (which might even be in a goroutine in future), I'd suggest calling Errorf instead.

cue/load/config.go
```diff
@@ -289,6 +312,9 @@ type Config struct {
 	// If an overlaid file does not exist in the host filesystem,
 	// the loader behaves as if the overlaid file exists with its contents,
 	// and that that all of its parent directories exist too.
+	//
+	// When FS is set, overlay paths must be absolute loadFS paths
+	// (that is, rooted at loadFsRoot).
```
rogpeppe, line 317: loadFsRoot is a private constant, so this wouldn't make much sense to the reader of the API docs. Also, in general, I'm not sure that it's right that the `@fs` prefix is exposed to the user. For example, if there are errors evaluating the CUE, the `@fs/` prefix will be exposed in the filenames in the errors, so if some tooling wants to do something with those files, it will have to first strip the prefix before it can use them in the filesystem.

Also, I wonder if Overlay is actually useful/required if we've got FS. Perhaps we should forbid supplying both at once.

cue/load/fs.go
```diff
@@ -318,7 +523,14 @@ func (fs *ioFS) absPathFromFSPath(name string) (string, error) {
 	if strings.ContainsAny(name, ":\\") {
 		return "", fmt.Errorf("invalid io/fs path %q", name)
 	}
-	return filepath.Join(fs.root, name), nil
+
```
rogpeppe, line 526: I suspect this logic isn't needed. This function (`absPathFromFSPath`) is used by `ioFS` to translate from fs path to the OS path used by the underlying `fileSystem` type, but ISTM that `ioFS` itself is mostly an unnecessary layer when the underlying files are provided by an `fs.FS`.

But... maybe it's not that easy! The various layers of caching and ReadCUEFile etc cause my brain to melt when I get to this point :)

rogpeppe (commented):
Thanks very much for having a bash at this!

I've done a first pass through this, and my main issue is that I'm not sure it's "right" to expose the `@fs` convention to the user, although I'm not sure what a better solution might look like. Clearly fs.FS doesn't have a notion of absolute paths, but I'm wondering if we should just go with the much-more-conventional `/` prefix to indicate an absolute path within the fs.FS.  That said, then the namespace within the FS overlaps with the namespace in the cache files, so perhaps we _do_ need some unconventional representation for files within the FS. If we do, then it should definitely be a well documented thing, because users and Go callers _will_ see those filenames.

I'm still thinking :)

mvdan:
@pskry just checking if you have seen Roger's review above? We are keen to move this forward if you're happy to continue with the contribution process.

rogpeppe:
Thanks again for this really great contribution! This acted as a really good jumping off point. After some reflection, we decided not to proceed directly from this PR, but to start the implementation from scratch, copying just some of the FS field doc comment from this PR, bearing in mind some of my review comments above. I've made a proposal [here](https://cue.gerrithub.io/c/cue-lang/proposal/+/1231971/1/designs/607-load-io-fs.md) and there's an implementation [here](https://cue.gerrithub.io/c/cue-lang/cue/+/1231627).

rogpeppe:
See discussion here: https://github.com/cue-lang/cue/discussions/4285

mvdan:
Roger's work has landed now (https://github.com/cue-lang/cue/commit/d267795de4d3d6e93f49ecd87e38440c9452d5e6), so I think we can close this PR. Thank you again for your help, @pskry!

PR #4262, opened 2026-02-12
internal/core/runtime: replace slice with sync.Map

jonjohnsonjr:
We have a workload that makes many calls to cue.Value.Path(), which spends most of its time inside IndexToString's RWMutex.Rlock() call.

This change attempts to optimize that by replacing the `labels` slice with a `sync.Map`.

From `sync.Map`'s godoc:

> The Map type is optimized for two common use cases:
> (1) when the entry for a given key is only ever written once but read
> many times, as in caches that only grow, ...
> In these two cases, use of a Map may significantly reduce lock
> contention compared to a Go map paired with a separate Mutex or RWMutex.

I believe this maps exactly to what IndexToString is doing, and indeed we spend about 50% less time inside IndexToString after this change.

### DIFF
--- a/internal/core/runtime/index.go
+++ b/internal/core/runtime/index.go
@@ -61,7 +61,7 @@ func (r *Runtime) Label(s string, isIdent bool) adt.Feature {
 // TODO: move to Runtime as fields.
 var (
 	labelMap = map[string]int{}
-	labels   = make([]string, 0, 1000)
+	labels   sync.Map
 	mutex    sync.RWMutex
 )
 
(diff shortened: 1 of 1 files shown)


mvdan:
Assigning to @rogpeppe given that he's been looking at concurrency issues with Marcel, and he's also been chatting with you recently :)

mvdan:
Roger's patch at https://cue.gerrithub.io/c/cue-lang/cue/+/1231635 does away with this code in favor of `unique.Handle`; we'll probably merge that once we can verify that we don't see any performance regressions.

Issue #1070, opened 2021-07-03
cmd/cue: tags across packages


cueckoo:
_Originally opened by @maurerbot in https://github.com/cuelang/cue/issues/1070_

`cue` seems to be unable to pick up tags across packages.

### What version of CUE are you using (`cue version`)?

<pre>
cue version v0.4.0 linux/amd64
</pre>

### Does this issue reproduce with the latest release?

Yes

### What did you do?

cue/cue.mod/module.cue
`module: my.tld`

cue/schema/base/base.cue
```
package base

#metadata: {
	name:         string @tag(name)
	description?: string
}

let Spec = {}

#base: {
	apiVersion: string
	kind:       string
	metadata:   #metadata
	spec?:      Spec
}
```

cue/schema/v1alpha1/test.cue
```
package v1alpha1

import "my.tld/schema/base"

let Spec = {
	keys?: [...{
		name: string
		pk:   =~"([a-zA-Z0-9_-]){8,256}"
	}]
} 

test: base.#base & {
	kind:       "test"
	apiVersion: "my.tld/v1alpha1"
	spec: Spec
}
```

```bash
cue eval ./schema/v1alpha1/... -t name="foo" -p v1alpha1 -e test --out=yaml
```

### What did you expect to see?

```yaml
apiVersion: "my.tld/v1alpha1"
kind:       test
metadata: 
    name: foo

spec: {...}
```

### What did you see instead?

`no tag for "name"`

Alternatively
```bash
cue eval ./schema/... -t name="foo"
#metadata: {
    name: "foo"
}
#base: {
    apiVersion: string
    kind:       string
    metadata: {
        name: "foo"
    }
}
// ---
test: {
    apiVersion: "my.tld/v1alpha1"
    kind:       "test"
    metadata: {
        name: string // this isn't picked up
    }
    spec: {}
}
```

## Other Questions
- Not sure if leveraging the lattice values as a means to define `Spec` is the right approach

cueckoo:
_Original reply by @maurerbot in https://github.com/cuelang/cue/issues/1070#issuecomment-866085374_

Adding `test:metadata:name: string @tag(name)` to `test.cue` seems to work but feels off to me.

cueckoo:
_Original reply by @verdverm in https://github.com/cuelang/cue/issues/1070#issuecomment-866288117_

Related is #1005

myitcv:
[...]

PR #4318, opened 2026-03-30
cue/load: tags across packages

bcachet:
tags marked with `scope=mod` are shared across packages of the same module
=> it allow to define some tag once in the module and reuse them in several packages/files

> [!NOTE]
> Fix issue reported in https://github.com/cue-lang/cue/issues/1070

### DIFF
--- a/cue/load/instances.go
+++ b/cue/load/instances.go
@@ -176,10 +176,16 @@ func Instances(args []string, c *Config) []*build.Instance {
 	// the same module. These tags are injected even when their package is
 	// not a root package but is merely imported.
 	if c.Module != "" {
-		visited := map[string]bool{}
+		// `capacity` is over-estimated because it counts shared and out-of-module
+		// deps, but it avoids rehashing as findModuleScopedTags adds entries.
+		capacity := len(a)
+		for _, p := range a {
+			capacity += len(p.Deps)
+		}
+		visited := make(map[string]struct{}, capacity)
 		// Seed visited with root package paths to avoid double-processing.
 		for _, p := range a {
-			visited[p.ImportPath] = true
+			visited[p.ImportPath] = struct{}{}
 		}
 		for _, p := range a {
 			tags, err := findModuleScopedTags(p, c.Module, visited)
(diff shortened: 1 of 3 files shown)


cue/load/instances.go
```diff
@@ -172,6 +172,24 @@ func Instances(args []string, c *Config) []*build.Instance {
 		tg.tags = append(tg.tags, tags...)
 	}
 
+	// Collect module-scoped tags (scope=mod) from transitive imports within
+	// the same module. These tags are injected even when their package is
+	// not a root package but is merely imported.
+	if c.Module != "" {
+		visited := map[string]bool{}
```
seh: Here, it looks like you know the _minimum_ required size of the map, but then you pass it down through `findModuleScopeTags` recursively, and continue adding to it. Can you estimate how many entries you'll wind up with?
```suggestion
		visited := make(map[string]bool, len(a))
```
Also, I'd rather see `map[string]struct{}` here, but you'd need to adjust your `findModuleScopeTags` signature to play along.

bcachet: Adapted accordingly in https://github.com/cue-lang/cue/pull/4318/commits/ec6d5f6a5e8bce8f1769b2be550680fe277bb09d

bcachet:
Hello

How should I proceed to get review/feedback on this PR ?

mvdan:
Hi - typically we prefer proposals or designs to be filed as issues so we can discuss and agree on the design before we look at the code. These days, the coding is the easier part :)

Is this implementing something that has been previously mentioned, like https://github.com/cue-lang/cue/issues/1070#issuecomment-1046281114? Or perhaps a variant of that? In any case, I would encourage you to summarize your design in a comment in that thread.

bcachet:
> Is this implementing something that has been previously mentioned, like [#1070 (comment)](https://github.com/cue-lang/cue/issues/1070#issuecomment-1046281114)? 

It is indeed implementing the idea proposed in https://github.com/cue-lang/cue/issues/1070#issuecomment-1046281114

I am adding a `scope` parameter to the tag definition. 
When set to `mod` the tag is shared across all packages of the module.
By default the `scope` parameter is set to `pkg` which does not change the current behavior

PR #4325, opened 2026-04-06
encoding/protobuf: respect lexical scoping rules

### LINKED ISSUE #4323: Protobuf parser does not respect lexical scoping rules

<pre>
$ cue version
cue version v0.16.0

CUE language version v0.16.0

Go version go1.26.0
      -buildmode exe
       -compiler gc
       -trimpath true
  DefaultGODEBUG cryptocustomrand=1,tlssecpmlkem=0,urlstrictcolons=0
     CGO_ENABLED 1
          GOARCH arm64
            GOOS darwin
         GOARM64 v8.0
             vcs git
    vcs.revision de47a5efb4a5ee1129a470e73717f59ac03ba535
        vcs.time 2026-03-03T13:21:57Z
    vcs.modified false
</pre>

Yes, and I can create a test that reproduces the problem.

### What did you do?

Tried to use `cue import` against https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/descriptor.proto 

A `.cue` file.

### What did you see instead?

```shell
error parsing protobuf: name "VisibilityFeature.DefaultSymbolVisibility" not found:
    ./src/google/protobuf/descriptor.proto:1188:12
```

This is due to a bug in the way that dotted references are resolved around [here](https://github.com/cue-lang/cue/blob/6e0f91cba468983cdf092251d7774795dc9d434c/encoding/protobuf/parse.go#L269); it always jumps to the top scope rather than going through lexical scopes for dotted paths. 

I have a fix for this I can put in a PR. Any reason not to do that?

As I was investigating this I also noticed that while the implementation correctly handles absolute (`.`-prefixed) message selectors, in the case of an lexically-dependent message selector (i.e. there's a top level message named `M` as well as a message named `M` in the current lexical scope) it's pretty hard to generate correct `.cue` files because we're going to generate a selector like `#M` which will always resolve to the current scope.

I don't know if this is a solvable problem, or if it would be worth solving (seems like an edge case). I couldn't figure out how to do it at all without a significant refactoring that touches complicated things (there's not enough information at the site where we could detect the problem to chain up the scopes and build a fully-qualified name (I think)), and it would probably be a breaking (or at least surprising) change to fully qualify all type names to avoid the problem anyway. But if someone can suggest a good solution I'd be happy to implement it.

SteveRuble:
This change modifies the protobuf converter to
respect lexical scoping when resolving relative
references. Previously, relative references were
resolved using the top-level scope, which meant
that references which were not fully-qualified
would fail to resolve. This change modifies the
logic so that it searches for the first element
of the reference name by iterating up the scope
chain until it finds a match.

This extends the existing .proto file used for
testing references to include a variety of
reference types, including relative,
fully-qualified, and ambiguous references.
It does NOT fix the existing bug where
non-fully-qualified references to shadowed
names end up always resolving to the shadowed
name in the generated CUE, because changing that
would require making the generated code opt in
to the experimental aliasv2 feature, which may
not be desirabled for all users.

This also fixes a bug in the proto generation
tests where the diff for golden files was
reversed, leading to confusing test failure
messages.

This work used LLM-powered tab completion,
but no other LLM tools.

Fixes cue-lang/cue#4323

### DIFF
--- a/encoding/protobuf/parse.go
+++ b/encoding/protobuf/parse.go
@@ -259,40 +259,52 @@ func (p *protoConverter) resolve(pos scanner.Position, name string, options []*p
 		return expr
 	}
 	if strings.HasPrefix(name, ".") {
-		return p.resolveTopScope(pos, name[1:], options)
+		// A leading dot indicates the path should be resolved in the top-level scope.
+		return p.resolveUsingScopes(pos, name[1:], p.scope[:1])
 	}
-	for _, scope := range slices.Backward(p.scope) {
-		if m, ok := scope[name]; ok {
-			return m.cue()
-		}
-	}
-	expr := p.resolveTopScope(pos, name, options)
+	// Non-absolute names need to be resolved using the nearest lexical scopes.
+	expr := p.resolveUsingScopes(pos, name, p.scope)
 	return expr
 }
 
-func (p *protoConverter) resolveTopScope(pos scanner.Position, name string, options []*proto.Option) ast.Expr {
+// resolveUsingScopes resolves a name using the nearest lexical scopes.
+// If the name is package-qualified, for this package, we trim the package prefix before searching.
+// If the name is dotted, we split the name and look for the first segment in the nearest lexical scope.
+// If the name is not found, we fail (panic with error).
+// scopes is a list of scopes to search, starting from the nearest scope;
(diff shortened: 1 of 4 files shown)


mvdan (approved):
Very nice, thank you. And thanks for your patience as I didn't get to this very quickly.

Issue #3133, opened 2024-05-08
encoding/openapi: full support for OpenAPI


jpluscplusm:
**Is your feature request related to a problem? Please describe.**

Users with existing OpenAPI documents can use the OpenAPI support initially tracked in #56 to work with their documents using the `cue` command, and the Go API.

However, as issues such as #1501 and #386 point out, users can't currently take an arbitrary OpenAPI document, `cue import` it, and then `cue export` their way back to the original document. Various fixed fields mentioned at https://swagger.io/specification/#fixed-fields (such as `paths`, `externalDocs`, and several keys under `components`) are currently ignored when CUE imports an OpenAPI document.

**Describe the solution you'd like**

I'd like CUE to implement full round-trip-ability from OpenAPI to CUE and back again; or to document those OpenAPI features/fields/elements for which round-tripping is unlikely to be achieved, so that OpenAPI users can make judgements about the degree of support that they require, based on their specific, individual OpenAPI use.

(Note that this doesn't ask for full round-trip-ability from arbitrary *CUE* to OpenAPI and back again)

**Describe alternatives you've considered**

n/a.

**Additional context**

This is being opened as part of the documentation work being performed in https://cuelang.org/cl/1192024.

rogpeppe:
The new whole-document OpenAPI implementation substantially addresses this: `ExtractV2`/`GenerateV2` operate on the entire document (not just `components.schemas`), preserving non-schema content verbatim and round-tripping OpenAPI 3.0 and 3.1 documents. It is available as a preview behind `CUE_EXPERIMENT=openapiv2` (introduced in v0.18.0). This issue will be closed once that experiment is enabled by default.

mvdan:
To be clear, the above is in master, and will be included in the upcoming v0.18.0-alpha.1 release. v0.18.0 will not be released for a little while yet.

rogpeppe:
Here's a short worked example of the new whole-document round-tripping

Enable the preview (available from v0.18.0-alpha.1):

```
export CUE_EXPERIMENT=openapiv2
```

Given an OpenAPI document `openapi.yaml`:

```yaml
openapi: 3.0.0
info:
  title: Example API
  version: "1.0"
paths:
  /people:
    get:
      responses:
        "200":
          description: a person
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Person'
components:
  schemas:
    Person:
      type: object
      properties:
        name: {type: string}
        age: {type: integer}
      required: [name]
```

[...]

PR #4331, opened 2026-04-17
encoding/openapi: handle default-wrapped numeric conjunctions

### LINKED ISSUE #4305: encoding/openapi: "unsupported op for number &" with ExpandReferences and default+bounded int
## Summary

`openapi.Gen` with `ExpandReferences: true` fails with `unsupported op for number &` when a definition contains a default value combined with a bounded integer conjunction.

## Minimal reproduction

```go
ctx := cuecontext.New()
v := ctx.CompileString(`#X: *20 | int & >=1 & <=100`)
_, err := openapi.Gen(v, &openapi.Generator{ExpandReferences: true})
// err: unsupported op for number &
```

## What works vs what doesn't

| Expression | ExpandReferences | Result |
|---|---|---|
| `#X: *20 \| int & >=1 & <=100` | `false` | OK |
| `#X: *20 \| int & >=1 & <=100` | `true` | **FAILS** |
| `#X: *20 \| int` | `true` | OK |
| `#X: int & >=1 & <=100` | `true` | OK |
| `#X: *20 \| >=1` | `true` | OK |
| `#X: null \| *20 \| >=1 & <=100` | `true` | OK |

The combination of a default value (`*20`), a disjunction (`|`), and a conjunction of type+bounds (`int & >=1 & <=100`) is required to trigger the bug.

## Where it fails

The error is produced at the default branch of the `number()` function:
https://github.com/cue-lang/cue/blob/66ea772b451293aac01f1bd2c610dfcf35af9fa3/encoding/openapi/build.go#L936-L942

`number()` handles individual constraint ops (`>=`, `<=`, `<`, `>`, `!=`, `CallOp`, `NoOp`) but has no case for `AndOp`. Normally conjunctions are decomposed before reaching `number()`, but with `ExpandReferences: true` the value arrives unsplit.

orvis98:
With ExpandReferences set, a numeric field that combines a bound conjunction with a default via disjunction, such as "int & >=1 | *1", would fail to generate an OpenAPI schema with "unsupported op for number &".

The value() dispatch pipeline relies on appendSplit to flatten a top-level conjunction into individual bound operands before handing each one to number(). When the value carries a default marker, cue.Value.Expr strips it and returns a NoOp wrapping the non-default arm. appendSplit's NoOp branch accepts that arm as a single conjunct without recursing, so a compound like "int & >=1" reaches number() with its AndOp intact and falls into the unsupported-op arm.

Handle AndOp in number() by recursing into each operand. The operands are simpler sub-expressions and hit the existing bound handlers, which merge minimum/maximum into the same schema node.

Fixes cue-lang/cue#4305.


mvdan:
Hi, thanks for the PR. With @rogpeppe we're currently working on entirely replacing the openapi encoding with a new implementation, so I don't think it's worth spending time on fixing bugs in the old code as the design and code are simply old and flawed. We'll update https://github.com/cue-lang/cue/issues/3133 once a version is ready to try out.

PR #4383, opened 2026-06-10
internal/core: fix Subsume nil-deref on BuiltinValidators with lazy args, allow bool != / == bounds

srnnkls:
Two related issues surfaced together while using cue.Value.Subsume as a
policy-matching primitive:

1. Subsume nil-derefs when a BuiltinValidator (e.g. list.MatchN) carries
   an argument containing a lazy reference such as a LetReference or a
   reference to a hidden helper field. The recursive subsume path creates
   a root Conjunct using s.ctx.Env(0) — which is a bare context with no
   binding for the referenced identifier — and the reference lookup walks
   off the end of the nil Environment chain.

   Fix: capture the construction-site Environment on BuiltinValidator and
   use it in subsume/value.go when building the validator's Conjunct, so
   lazy args resolve against the scope in which the validator was written.

2. BoundExpr.evaluate rejected != / == bounds on BoolKind values with
   "invalid value ... for bound !=", leaving the bound as bottom. This is
   asymmetric with null / struct / list, which do accept == / != bounds.

   Fix: add BoolKind to the set of kinds that accept == / != bounds in
   BoundExpr.evaluate.

Regression tests:
- cue/types_test.go: TestSubsume cases for let-reference interpolation,
  hidden-field interpolation inside a regex bound.
- internal/core/subsume/structural_test.go: bool != / == bound subsumption
  cases (symmetric with the existing numeric and string bound tests).

Signed-off-by: Sören Nikolaus <soeren@code17.io>

### DIFF
--- a/cue/types_test.go
+++ b/cue/types_test.go
@@ -2220,6 +2220,39 @@ func TestSubsume(t *testing.T) {
 		pathA: a,
 		pathB: b,
 		want:  false,
+	}, {
+		value: `
+			import "list"
+			let h = "abc"
+			a: list.MatchN(>0, =~"\(h)")
+			b: ["abc"]
+			`,
+		pathA:   a,
+		pathB:   b,
+		options: []cue.Option{cue.Final()},
+		want:    true,
+	}, {
+		value: `
+			import "list"
+			let h = "abc"
+			a: list.MatchN(>0, =~"\(h)")
+			b: ["xyz"]
+			`,
+		pathA:   a,
+		pathB:   b,
+		options: []cue.Option{cue.Final()},
+		want:    false,
+	}, {
+		value: `
+			import "list"
+			_h: "abc"
+			a: list.MatchN(>0, =~"\(_h)")
+			b: ["abc"]
+			`,
+		pathA:   a,
+		pathB:   b,
+		options: []cue.Option{cue.Final()},
+		want:    true,
 	}}
 	for _, tc := range testCases {
 		cuetdtest.FullMatrix.Run(t, tc.value, func(t *testing.T, m *cuetdtest.M) {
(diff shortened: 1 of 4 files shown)


mvdan:
Could you please raise two bug reports for these? It doesn't seem wise to review code before there's agreement on what the problem is. For example, the boolean change doesn't make much sense to me, because strings and numbers can also be compared.

srnnkls:
Thank you! Sure, will do.

srnnkls:
Filed the two issues: #4387 (Subsume `nil-deref`) and #4388 (`bool` bounds).

On the boolean change... Agreed that strings and numbers can be compared, and that's actually the point: they already accept `==`/`!=` bounds (`!=5`, `!="foo"`), and so do `null`, `structs`, and `lists` (the spec has `!=null & 1 // 1` and `!=[1, 2] & [1] // [1]` as examples). `Bool` is currently the only kind that rejects them, even though the spec defines `==`/`!=` for any concrete operands. Ordered bounds like `<true` stay rejected, of course. There's a full kind × operator breakdown in #4388.

mvdan:
Could you split this into two patches? The fixes seem separate, so it would be best to review and test them separately.

If you can, please send them directly to Gerrit as well: https://github.com/cue-lang/cue/blob/master/CONTRIBUTING.md#preparing-for-gerrithub-cl-contributions

srnnkls:
Done — split into two and sent to Gerrit:

- Subsume nil-deref (#4387) → https://cue.gerrithub.io/c/cue-lang/cue/+/1239901
- bool `!=` / `==` bounds (#4388) → https://cue.gerrithub.io/c/cue-lang/cue/+/1239902

Happy to close this PR in favour of the CLs — whichever you prefer.

mvdan:
Thanks! Yes, I'll look at the patches on Gerrit instead.

PR #4417, opened 2026-07-01
encoding/protobuf: resolve unqualified references to package-less proto types

### LINKED ISSUE #4416: encoding/protobuf: unqualified references to types in a package-less .proto fail to resolve

<pre>
$ cue version
cue version v0.18.0-0.dev.0.20260630152551-ac9a510b4692

CUE language version v0.18.0
Go version go1.26.0
</pre>

(Also reproduces on the released `v0.16.0`.)

Yes. Confirmed on `v0.16.0` and on current `master` (`ac9a510b4`).

### What did you do?

`cue import proto` cannot resolve an unqualified reference to a message/enum
that is declared in a `.proto` file with **no `package` declaration**, when
that reference comes from a file in a different, named package. Types in a
package-less file live in the global proto namespace and are reachable
unqualified; `protoc` resolves this fine. It is standard proto name
resolution.

Minimal reproduction, two files in the same directory.

`global_scope.proto` — no `package`, so its types live in the global scope:

```proto
syntax = "proto3";
option go_package = "example.com/global;globalpb";
message GlobalMessage { int64 value = 1; }
```

`user.proto` — a named package referencing `GlobalMessage` unqualified:

```proto
syntax = "proto3";
package myns;
import "global_scope.proto";
message User { GlobalMessage m = 1; }
```

Then:

```
$ protoc --proto_path=. -o /dev/null user.proto   # reference impl: succeeds (exit 0)
$ cue import proto -I . user.proto                 # cue: fails
error parsing protobuf: name "GlobalMessage" not found:
    ./user.proto:4:16
```

`cue import proto` resolves `GlobalMessage` to the type declared in the
package-less file (matching `protoc`), and generates the CUE for `user.proto`
with a qualified reference and an import of the global file's package.

### What did you see instead?

```
error parsing protobuf: name "GlobalMessage" not found:
    ./user.proto:4:16
```

### Root cause

In `encoding/protobuf/parse.go`, `protoConverter.protoPkg` is set only from an
explicit `proto.Package` element, so a package-less file leaves
`protoPkg == ""`.

`doImport` then registers the imported file's symbols into the importer's top
scope:

```go
pkgNamespace := strings.Split(imp.protoPkg, ".")
...
ref := k
if len(pkgNamespace) > 0 {
    ref = strings.Join(append(pkgNamespace, k), ".")
}
...
p.scope[0][ref] = mapping{f, pkg}
```

`strings.Split("", ".")` returns `[""]` — a slice of length **1**, not an
empty slice. So for a package-less import `len(pkgNamespace) > 0` is true, and
the symbol is registered under a key with a spurious **leading dot**, e.g.
`.GlobalMessage` instead of `GlobalMessage`.

Resolution (`resolve` → `resolveUsingScopes`) looks the symbol up by its bare
name `GlobalMessage`. A leading `.` in a reference is stripped by `resolve`
before the lookup, and `resolveUsingScopes` never re-adds one, so neither the
unqualified form (`GlobalMessage`) nor the fully-qualified global form
(`.GlobalMessage`) ever matches the registered `.GlobalMessage` key. The symbol
is unreachable, and resolution falls through to `name %q not found`.

### Why no existing test caught it

Every `.proto` under `encoding/protobuf/testdata` declares a `package`, so the
empty-`protoPkg` path is never exercised.

### Notes

Adjacent name-resolution work in this file is recent (#4323 / #4325 reworked
`resolveUsingScopes`; #4035 / #4036 handled fully-qualified names within a
package), but none touched the `doImport` registration loop, so this
package-less case remains.

stefan-toubia:
Fixes #4416.

`cue import proto` failed to resolve an unqualified reference to a type
declared in a `.proto` file with no `package` declaration when referenced from
another package, because `strings.Split("", ".")` returns `[""]` and the
symbol was registered under a leading-dot key (`.Foo`) that no reference
resolves to. `protoc` accepts the same input. See the linked issue for a
minimal reproduction.

Adds a `testdata` fixture (a package-less proto imported and referenced
unqualified from a named package); no existing fixture omitted a `package`
clause, which is why this was uncaught.

### DIFF
--- a/encoding/protobuf/parse.go
+++ b/encoding/protobuf/parse.go
@@ -388,8 +388,13 @@ func (p *protoConverter) doImport(v *proto.Import) error {
 		fail(v.Position, err)
 	}
 
-	pkgNamespace := strings.Split(imp.protoPkg, ".")
-	curNamespace := strings.Split(p.protoPkg, ".")
+	var pkgNamespace, curNamespace []string
+	if imp.protoPkg != "" {
+		pkgNamespace = strings.Split(imp.protoPkg, ".")
+	}
+	if p.protoPkg != "" {
+		curNamespace = strings.Split(p.protoPkg, ".")
+	}
 	for {
 		for k := range imp.symbols {
 			ref := k
--- a/encoding/protobuf/protobuf_test.go
+++ b/encoding/protobuf/protobuf_test.go
@@ -38,6 +38,7 @@ func TestExtractDefinitions(t *testing.T) {
 		"mixer/v1/config/client/client_config.proto",
 		"other/trailcomment.proto",
 		"other/full_references.proto",
+		"other/package_less_import.proto",
 	}
 	for _, file := range testCases {
 		t.Run(file, func(t *testing.T) {
--- a/encoding/protobuf/testdata/istio.io/api/other/global_scope.proto
+++ b/encoding/protobuf/testdata/istio.io/api/other/global_scope.proto
@@ -1,0 +1,10 @@
+syntax = "proto3";
+
+// This file intentionally declares no `package`: its messages live in the
+// global proto namespace. It mirrors well-known-type style files that are
+// imported and referenced unqualified from other, named packages.
+option go_package = "istio.io/api/other/global_scope;global_scope";
+
+message GlobalMessage {
+  int64 value = 1;
+}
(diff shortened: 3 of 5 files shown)


mvdan (approved):
Thanks! LGTM, no notes.

Issue #2530, opened 2023-08-08
encoding/openapi: panic: unreachable HiddenLabel + IndexLabel


mvdan:
### What version of CUE are you using (`cue version`)?

<pre>
$ cue version
cue version v0.0.0-20230802143915-da3ee7dc7b27

go version devel go1.22-088f953fd0 Mon Aug 7 06:58:18 2023 +0000
      -buildmode exe
       -compiler gc
  DefaultGODEBUG panicnil=1
     CGO_ENABLED 1
          GOARCH amd64
            GOOS linux
         GOAMD64 v3
             vcs git
    vcs.revision da3ee7dc7b27de85bbfb0a4104dfb22bcb18ef68
        vcs.time 2023-08-02T14:39:15Z
    vcs.modified false
</pre>

### Does this issue reproduce with the latest stable release?

Yes, verified with v0.5.0.

### What did you do?

```
go mod tidy
go run .
-- main.go --
package main

import (
	"bytes"
	"encoding/json"
	"fmt"
	"log"

	"cuelang.org/go/cue"
	"cuelang.org/go/cue/cuecontext"
	"cuelang.org/go/cue/load"
	"cuelang.org/go/encoding/openapi"
)

func genOpenAPI(inst cue.InstanceOrValue) ([]byte, error) {
    b, err := openapi.Gen(inst, nil)
    if err != nil {
        return nil, err
    }

    var out bytes.Buffer
    err = json.Indent(&out, b, "", "   ")
    if err != nil {
        return nil, err
    }
    
    return out.Bytes(), nil
}

func main() {
	inst := load.Instances([]string{"."}, &load.Config{Dir: "mod"})[0]
	if err := inst.Err; err != nil {
		log.Fatal(err)
	}
	val := cuecontext.New().BuildInstance(inst)
	if err := val.Err(); err != nil {
		log.Fatal(err)
	}

    b, err := genOpenAPI(val)
    if err != nil {
		log.Fatal(err)
    }
	fmt.Println(string(b))
}

-- mod/cue.mod/module.cue --
module: "mod.test"
-- mod/x.cue --
package x

#Def: {
	_Another
	barRadius: float64
}

_Another: {
	barRadius: < 2.5
}

-- go.mod --
module test

go 1.20

replace cuelang.org/go => /home/mvdan/src/cue/cue

require cuelang.org/go v0.0.0-00010101000000-000000000000
-- go.sum --
```

### What did you expect to see?

No panic.

### What did you see instead?

[...]

PR #4424, opened 2026-07-02
update encoding/openapi error instead of panic on unsupported selectors

robmyersrobmyers:
fix: update encoding/openapi to fail with an error instead of a panic when unsupported selectors are found.

- Updates selectorLabel() to return ok so it can return false instead of a panic.
- Moves failf() logic from builder to buildcontext so that failf() can be accessed from either.
- Updates unsupported() in the builder to check if selectorLabel() returns false and uses failf() to return a helpful error message.

Replaces 'panic: unreachable IndexLabel [recovered, repanicked]' with '#A: cannot create OpenAPI reference for #List[0]: unsupported path element 0' when running 'cue def schema.cue -o openapi+yaml:schema.yaml'.

This is not a cue v0.17 regression; cue v0.16.1 exhibits the same panic.

Improves #2530 so it at least doesn't panic.  The error message has proven useful in my testing.

### DIFF
--- a/encoding/openapi/build.go
+++ b/encoding/openapi/build.go
@@ -181,12 +181,7 @@ func (c *buildContext) isInternal(sel cue.Selector) bool {
 }
 
 func (b *builder) failf(v cue.Value, format string, args ...interface{}) {
-	panic(&openapiError{
-		errors.NewMessagef(format, args...),
-		v.Err(),
-		cue.MakePath(b.ctx.path...),
-		v.Pos(),
-	})
+	b.ctx.failf(v, format, args...)
 }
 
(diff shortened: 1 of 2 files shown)


mvdan:
Hi, thanks for the PR. With @rogpeppe we're currently working on entirely replacing the openapi encoding with a new implementation, so I don't think it's worth spending time on fixing bugs in the old code as the design and code are simply old and flawed. We'll update https://github.com/cue-lang/cue/issues/3133 once a version is ready to try out.

mvdan:
Closing per the above. We will share some news next week.

PR #4433, opened 2026-07-12
modernize Go code via go fix ./...

udf2457:
Use `go fix ./...` to modernise code, e.g replace `interface{}` with `any`

```sh
$ go test ./cue/format -run TestFiles -count=1
ok      cuelang.org/go/cue/format       0.009s
$ go test ./cue/format ./internal/pretty
ok      cuelang.org/go/cue/format       (cached)
ok      cuelang.org/go/internal/pretty  (cached)
```


mvdan:
Thanks. We are aware of `go fix`, and we have been applying these fixes piecemeal over time to reduce the amount of conflicts we suffer. We will continue to apply them slowly.

Issue #4371, opened 2026-05-29
cmd/cue: major version defaults resolved relative to main module, not importing module


rogpeppe:
### What version of CUE are you using (`cue version`)?

<pre>
74211d77f0538ac9b88fb937b9d1e03712769f72</pre>
</pre>

### Does this issue reproduce with the latest stable release?

Yes (v0.16.1)

### What did you do?

When a dependency module declares its own default major version for a
transitive dependency, importing that transitive dependency without an
explicit major version qualifier resolves the default relative to the
main module's `module.cue`, not relative to the importing module's own
`module.cue`.

Reproduction:

```txtar
exec cue export .
cmp stdout want-eval.txt

-- cue.mod/module.cue --
module: "example.com/main@v0"
language: version: "v0.9.0"

deps: {
	"example.com/mod@v0": {
		default: true
		v: "v0.0.0"
	}
	"example.com/mod@v1": {
		v: "v1.0.0"
	}
	"example.com/other": {
		default: true
		v: "v0.0.0"
	}
}
-- main.cue --
package main

import (
	"example.com/mod"
	"example.com/other"
)
"mod": mod.value
"other": other.value
-- _registry/example.com_mod_v0.0.0/cue.mod/module.cue --
module: "example.com/mod@v0"
language: version: "v0.9.0"
-- _registry/example.com_mod_v0.0.0/mod.cue --
package mod

value: "v0"

-- _registry/example.com_mod_v1.0.0/cue.mod/module.cue --
module: "example.com/mod@v1"
language: version: "v0.9.0"
-- _registry/example.com_mod_v1.0.0/mod.cue --
package mod

value: "v1"

-- _registry/example.com_other_v0.0.0/cue.mod/module.cue --
module: "example.com/other@v0"
language: version: "v0.9.0"

deps: {
	"example.com/mod@v1": {
		default: true
		v: "v1.0.0"
	}
}
-- _registry/example.com_other_v0.0.0/other.cue --
package other

import "example.com/mod"

value: mod.value

-- want-eval.txt --
{
    "mod": "v0",
    "other": "v1"
}
```

### What did you expect to see?

The `example.com/other` module declares `example.com/mod@v1` as its default
for `example.com/mod`. When `other.cue` imports `"example.com/mod"` without
a major version qualifier, it should resolve to `example.com/mod@v1` based
on its own module's default, producing:

```json
{
    "mod": "v0",
    "other": "v1"
}
```

### What did you see instead?

[...]

Issue #3162, opened 2024-05-21
cue/load: using qualifier that is the same as final path element causes import failure


rogpeppe:
### What version of CUE are you using (`cue version`)?

<pre>
a90aec5c611bda3d73dfe484d4794e07499ff2ae
</pre>

### Does this issue reproduce with the latest stable release?

Yes (at v0.8.2)

### What did you do?

```
# old
env CUE_EXPERIMENT=modules=0
exec cue vet -c .

# new
env CUE_EXPERIMENT=modules
exec cue vet -c .

-- cue.mod/module.cue --
module: "github.com/foo/bar"
language: {
	version: "v0.9.0"
}
-- bar.cue --
package bar

import xxx "github.com/foo/bar/abc:abc"

z: xxx.x & 123

-- abc/baz.cue --
package abc

x: 123
```

### What did you expect to see?

A passing test

### What did you see instead?

```
# old (0.023s)
# new (0.011s)
> env CUE_EXPERIMENT=modules
> exec cue vet -c .
[stderr]
github.com/foo/bar@v0: import failed: no dependency found for package "github.com/foo/bar/abc":
    ./bar.cue:3:8
[exit status 1]
FAIL: /tmp/testscript3893947838/repro.txtar/script.txtar:7: unexpected command failure
```

myitcv:
It was ambiguous which vet call had failed, so I've updated the original description. 

@rogpeppe - to confirm we will fix this as part of the work in #3155?

rogpeppe:
> It was ambiguous which vet call had failed, so I've updated the original description.
> 
> @rogpeppe - to confirm we will fix this as part of the work in #3155?

The fix is actually independent of that, but I have a fix and it will land at or around the same time.

PR #4445, opened 2026-07-17
internal/mod/modpkgload: canonicalize resolved import lookups

cjorge-graphops:
Commit b24b39cf made unversioned imports resolve against the importing external module's defaults to address cue-lang/cue#4371. It records per-package rewrites to explicit major-version paths so cue/build's path-keyed cache does not conflate different resolutions.

That handoff assumed the module scanner and builder saw the same spelling. AllImports canonicalizes redundant qualifiers, while the build callback receives the original AST spelling. An entry recorded under example.com/mod was therefore queried as example.com/mod:mod. Depending on whether the main module had already loaded the unversioned identity, this either reused the wrong major or failed with no dependency found even though cue mod tidy accepted the graph.

Canonicalize the source spelling for the resolvedImports lookup only, returning it unchanged when no rewrite exists. This keeps the map canonical rewrite input to canonical rewrite result, preserves the canonical package-identity behavior from 26db19be and cue-lang/cue#3162, and avoids globally changing observable build.Instance.ImportPath values.

Replacement plumbing added in 74d78c17 can change whether a qualifier is redundant after rewriting a module path. Canonicalize rewritten replacement paths too, preventing noncanonical intermediate keys from weakening the invariant.

Cover the boundary and replacement cases directly, and add command regressions for both the hard miss and silent wrong-major reuse. The hard-miss fixture also verifies cue mod tidy --check succeeds.

### DIFF
--- a/cmd/cue/cmd/testdata/script/module-major-version-default-qualified-collision.txtar
+++ b/cmd/cue/cmd/testdata/script/module-major-version-default-qualified-collision.txtar
@@ -1,0 +1,72 @@
+# A qualified import in an external module must use that module's default
+# major version even when the main module has already loaded the same
+# unversioned import under a different default.
+
+exec cue export .
+cmp stdout want-eval.txt
+
+-- cue.mod/module.cue --
+module: "example.com/main@v0"
+language: version: "v0.9.0"
+
+deps: {
+	"example.com/mod@v0": {
+		default: true
+		v: "v0.0.0"
+	}
+	"example.com/mod@v1": {
+		v: "v1.0.0"
+	}
+	"example.com/other": {
+		default: true
+		v: "v0.0.0"
+	}
+}
+-- main.cue --
+package main
+
+import (
+	"example.com/mod"
+	"example.com/other"
+)
+"mod": mod.value
+"other": other.value
+
+-- _registry/example.com_mod_v0.0.0/cue.mod/module.cue --
+module: "example.com/mod@v0"
+language: version: "v0.9.0"
+-- _registry/example.com_mod_v0.0.0/mod.cue --
+package mod
+
+value: "v0"
+
+-- _registry/example.com_mod_v1.0.0/cue.mod/module.cue --
+module: "example.com/mod@v1"
+language: version: "v0.9.0"
+-- _registry/example.com_mod_v1.0.0/mod.cue --
+package mod
+
+value: "v1"
+
+-- _registry/example.com_other_v0.0.0/cue.mod/module.cue --
+module: "example.com/other@v0"
+language: version: "v0.9.0"
+
+deps: {
+	"example.com/mod@v1": {
+		default: true
+		v: "v1.0.0"
+	}
+}
+-- _registry/example.com_other_v0.0.0/other.cue --
+package other
+
+import "example.com/mod:mod"
+
+value: mod.value
+
+-- want-eval.txt --
+{
+    "mod": "v0",
+    "other": "v1"
+}
--- a/cmd/cue/cmd/testdata/script/module-major-version-default-qualified.txtar
+++ b/cmd/cue/cmd/testdata/script/module-major-version-default-qualified.txtar
@@ -1,0 +1,58 @@
+# A redundant package qualifier must not prevent an external module's
+# unversioned import from using that module's default major version.
+# The main module deliberately does not import example.com/mod itself, so a
+# missed contextual rewrite cannot fall back to an unversioned package entry.
+
+exec cue mod tidy --check
+exec cue export .
+cmp stdout want-eval.txt
+
+-- cue.mod/module.cue --
+module: "example.com/main@v0"
+language: version: "v0.9.0"
+
+deps: {
+	"example.com/mod@v1": {
+		v: "v1.0.0"
+	}
+	"example.com/other": {
+		default: true
+		v: "v0.0.0"
+	}
+}
+-- main.cue --
+package main
+
+import "example.com/other"
+
+value: other.value
+
+-- _registry/example.com_mod_v1.0.0/cue.mod/module.cue --
+module: "example.com/mod@v1"
+language: version: "v0.9.0"
+-- _registry/example.com_mod_v1.0.0/mod.cue --
+package mod
+
+value: "v1"
+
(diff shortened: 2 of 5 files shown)


cjorge-graphops:
This is AI generated, but the failure is real (v0.17.0/v0.17.1 broke my stuff), and tried to make sure it's sensible.

Here's a more in depth discussion of the issue:

# Correct the import-resolution handoff introduced by b24b39cf

  ## Full rationale

  Before b24b39cf, CUE effectively gave an unversioned import one global meaning: the default major selected by the main module.

  Issue #4371 demonstrated why that was wrong. Given:

  main module:
      example.com/mod defaults to @v0

  external module "other":
      example.com/mod defaults to @v1

  both modules can contain:

  import "example.com/mod"

  but those imports must resolve to different package identities.

  Commit b24b39cf correctly changed module resolution to use the importing external module’s requirements. It also recognized that cue/build caches imports by path string: leaving both imports as example.com/mod would cause the build cache to conflate
  v0 and v1. Therefore, the external module’s import had to become an explicit identity such as example.com/mod@v1.

  The commit introduced resolvedImports to carry that decision from module scanning into the later build pass:

  module scanner determines:
      example.com/mod → example.com/mod@v1

  cue/build later rereads the AST:
      callback replays that rewrite before cache lookup

  That architecture is sound.

  ## The mistake

  The two passes have different input contracts:

  - modimports.AllImports returns syntactically canonical paths.
  - cue/build passes the callback the spelling extracted directly from the AST.

  For a normally written import, those strings are identical, which is all the commit’s test covered. They differ when a redundant qualifier is present:

  source spelling:       example.com/mod:mod
  AllImports result:     example.com/mod
  resolved identity:     example.com/mod@v1

  The commit consequently creates:

  resolvedImports["example.com/mod"] = "example.com/mod@v1"

  but later performs:

  resolvedImports["example.com/mod:mod"]

  That lookup misses.

  The “raw import path” comment appears to have used “raw” to mean “before major-version resolution.” It did not account for the separate distinction between literal source spelling and syntactically canonical spelling.

  This explains both observed outcomes:

  - If no other resolution is available, cue/build reports “no dependency found.”
  - If the main module already loaded the unversioned path under another default major, path-based deduplication can silently reuse the wrong package.
  - cue mod tidy --check still passes because the module-scanning pass saw the dependency correctly; the failure occurs only when replaying that resolution in the build pass.

  ## Correct representation and fix

  resolvedImports should have one strict representation:

  canonical path before contextual resolution
      →
  canonical path after contextual resolution

  Raw source spellings must not be stored as alternative keys.

  The boundary method should be:

  func (pkg *Package) CanonicalImportPath(sourcePath string) string {
      lookupPath := ast.ParseImportPath(sourcePath).Canonical().String()
      if resolvedPath, ok := pkg.resolvedImports[lookupPath]; ok {
          return resolvedPath
      }
      return sourcePath
  }

  The original spelling is returned on a miss because this method is a conditional contextual-rewrite hook. Canonicalization is used to query package identity; it should not incidentally rewrite imports for which modpkgload made no resolution
  decision.

  There should be no exact-raw-key fallback. Such a fallback would make the immediate reproducer pass while implicitly permitting two key representations.

  ## Replacement invariant

  resolvedImports now has two writers:

  - Per-module default-major resolution, which already produces ip.Canonical().String().
  - Module replacement rewriting, which currently returns parts.String() after changing the path.

  A replacement can change whether a qualifier is redundant. Therefore, replacement output must also use:

  parts.Canonical().String()

  when a rewrite occurs. Unmatched paths remain unchanged.

  This prevents replacement handling from introducing the mixed representation that the boundary fix is intended to eliminate.

  ## Change size and blast radius

  The expected production change is very small:

  - internal/mod/modpkgload/pkgload.go
      - Canonicalize the callback argument before the map lookup.
      - Correct the map and method comments.

  - internal/mod/modpkgload/replace.go
      - Canonicalize the result of an actual replacement rewrite.

  This is only a handful of executable lines across two internal files. No changes should be needed in cue/build, cue/load, module selection, registry access, MVS, or module-file parsing.

  Behavioral scope is narrow:

  - External-module imports that have contextual resolution entries.
  - Imports affected by a configured module replacement.
  - Specifically, spellings whose qualifier differs from their canonical form.

  Unaffected behavior includes:

  - Main-module imports without replacements.
  - Standard-library imports.
  - Already-canonical imports.
  - Explicitly versioned imports without a contextual rewrite.
  - Dependency selection and registry fetching.
  - Unmapped imports, which retain their original spelling.

  The code path is identity-sensitive—import paths drive caching, deduplication, cycle detection, and hidden-field namespaces—so the semantic risk is more important than the line count. However, the fix does not compute any new resolution; it only
  ensures cue/build receives the resolution already computed by modpkgload.

  The replacement change may alter an observable rewritten path only when a replacement makes an explicit qualifier redundant. That is the canonical spelling of the same package identity and matches the method’s existing contract.

  There are no public API or data-format changes, making the patch appropriate for v0.17.x backporting.

  ## Regression proof

  Add focused internal tests covering:

  - Redundant source spelling finds a canonical map key.
  - Already-canonical spelling produces the same resolved identity.
  - A genuinely nonredundant qualifier remains significant.
  - An unmapped redundant spelling is returned exactly unchanged.
  - Test fixtures never place noncanonical keys in resolvedImports.

  Add replacement tests covering:

  - Replacement makes a qualifier redundant: canonical result drops it.
  - Replacement leaves a qualifier necessary: result retains it.
  - No matching replacement: input is unchanged.

  Add end-to-end tests for both consequences of the original bug:

  - Hard failure: a qualified external import with a declared dependency no longer reports “no dependency found.”
  - Silent collision: main and external modules select different default majors for the same qualified import, and each receives the correct version.

mvdan:
Reiterating what I shared on Slack - please file a bug report following the template, to make sure we're tracking this and understand the problem before we look at any fix.

cjorge-graphops:
Fixes https://github.com/cue-lang/cue/issues/4446

mvdan:
Thanks! Replied there: https://github.com/cue-lang/cue/issues/4446#issuecomment-5023588373

If that sounds good to you, then I believe we can close this PR, and we should merge the stack on Gerrit today or tomorrow.

cjorge-graphops:
Closing now as discussed :saluting_face: :rocket:

